PR #1, opened 2018-08-02
Issue fix001

zeusintuivo:
+- added missing newline
+- added key to for loop
+- added key for loop
+- updated package by installing


src/views/Profile.vue
```diff
@@ -50,7 +50,9 @@
       <header>
         <h2>Relations</h2>
       </header>
-      <Person v-for="person in todo.team.members" v-bind="person" />
+      <Person v-for="person in todo.team.members"
+              :key="person"
+              v-bind="person" />
```
hidde: Like the use of multiple lines… note to self: we should definitely this in other places, too.

PR #17, opened 2018-09-07
Add Dockerfile and use proxy for dev.

### LINKED ISSUE #15: Dockerize
Have a Dockerfile to create an artifact serving the new front-end.

Most likely we should just serve a prod build from nginx.

fiji-flo:
This resolves #15.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,21 @@
+.DS_Store
+node_modules
+/dist
+
+# local env files
+.env.local
+.env.*.local
+
+# Log files
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# Editor directories and files
+.idea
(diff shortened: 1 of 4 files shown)


.dockerignore
```diff
@@ -0,0 +1,21 @@
...
+yarn-error.log*
+
+# Editor directories and files
+.idea
+.vscode
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw*
```
hidde: this is equivalent to the `.gitnore`, can we be smarter here? This may very well qualify as premature optimisation and there's probably no need as we will likely hardly touch these files, but just wondered seeing this.

fiji-flo: Yeah, I would iterate on this. I'll add an issue.

PR #19, opened 2018-09-10
Update hris field names.

fiji-flo:


src/queries/profile.js
```diff
@@ -6,13 +6,13 @@ const PROFILE_QUERY = gql`query ($userId: String!) {
...
-          wprDeskNumber,
-          costCenter,
+          Team,
+          Entity,
+          LocationDescription,
+          TimeZone,
+          WorkerType,
+          WPRDeskNumber,
+          CostCenter,
         }
```
hidde: Curious about the reason for this change… all other properties in this file are camelCase and with UpperCamelCase they no longer match GraphQL query responses from the profile faker? (I am likely missing something)

akatsoulas: @hidde you are correct. HRIS data have a special formatting and they need to be like this. In order to have a consistent name formatting I opened this PR [0] which basically allows us to query GraphQL in a consistent way regardless of the data format (cc @fiji-flo )

[0] https://github.com/mozilla-iam/iam-profile-faker/pull/21

akatsoulas: This also means that when we merge this other PR, this is no longer needed.

PR #26, opened 2018-09-23
[DNM] Simple orgchart list.

fiji-flo:


.eslintrc.js
```diff
@@ -1,18 +1,18 @@
 module.exports = {
   root: true,
   env: {
-    node: true
+    node: true,
```
hidde: Yay, thanks for the commas!  💯

src/components/OrgNode.vue
```diff
@@ -0,0 +1,71 @@
...
+};
+</script>
+
+<style>
+  .org-node label {
+    width: 100%;
+  }
+
+  .org-node label:hover {
+      background-color: lightgray;
```
hidde: for colours, can we stick to the vars defined in the App component? So this would be: `var(--lightGrey)` (sidenote: we may at some point want to review the greys)

src/components/OrgNode.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+  <div class="org-node">
+    <input :id="`${prefix}-${id}`" type="checkbox">
```
hidde: Maybe this should move inside the `li`, because `ul` can only contain `li`s as children?

hidde: My bad, the `div` wrapping this would need to move inside the `li`

src/components/OrgNode.vue
```diff
@@ -0,0 +1,71 @@
...
+    <li v-bind:class="{ branch: children.length }">
```
hidde: To stick to the BEM convention, could we use `org-node__branch` here?

src/components/OrgNode.vue
```diff
@@ -0,0 +1,71 @@
...
+
+  .org-node input {
+      position: absolute;
+      opacity: 0;
+      cursor: pointer;
+  }
+
+  .org-node li {
+      list-style: disc;
+      display: list-item;
```
hidde: is this not the default `display` value for `li`'s?

src/components/OrgNode.vue
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  .org-node input {
+      position: absolute;
+      opacity: 0;
+      cursor: pointer;
+  }
+
+  .org-node li {
+      list-style: disc;
```
hidde: is this not the default `list-style`?

src/components/OrgRoot.vue
```diff
@@ -0,0 +1,33 @@
...
+  props: {
+    roots: Array,
+  },
+};
+</script>
+
+<style>
+  .org-root {
+    width: 80%;
+    margin: 0px auto;
```
hidde: for consistency can we use `0` instead of `0px`?

src/views/PageOrgchart.vue
```diff
@@ -0,0 +1,67 @@
+<template>
+  <div class="orgchart">
+    <LoadingSpinner v-if="loading"></LoadingSpinner>
+    <div v-else-if="error" class="error">
+        <Error>
```
hidde: (nitpick alert)  this has 4 spaces instead of  2?

src/components/OrgNode.vue
```diff
@@ -0,0 +1,71 @@
...
+  .org-node li.branch {
+      list-style-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAkAAAAJCAYAAADgkQYQAAAApUlEQVQYlWNonzS/vHPygv+4cPuk+eUMj5+/KZ86bxVWBVPnrfr/+PmbcgYGBgaGE+evvMOm6MDRs+8YYODJ81ehC1ZuRlGwYOXm/09evk5jQAbX7zw4g6zoxLmrZxjQwcOXL5W2bt37v3Pygv9rNu/9//TlSxcMRQwMDAx37j+e2T9j6f8rN+6twqqAgYGB4eHDh4LX7jx49/DlSyWcimDWoosBAHcwjGsLIU2wAAAAAElFTkSuQmCC);
+  }
+
+
+  .org-node label + .toggle + ul {
+      display: inline-block;
+  }
+
+  .org-node input:checked + li ul {
```
hidde: Do we need `li` _and_ `ul` for reasons? Otherwise this could be shorter

fiji-flo:
I think I addressed all concerns. Thanks for pointing them out.

PR #30, opened 2018-10-03
Style org chart, add UX

hidde:
Updates in this PR: 

* Moved from checkbox pattern to using `<ShowMore>` component
* Introduced Vuex to have a global state. First use cases: storing what profile is being previewed in the org chart; storing whether profile is previewed on small or large screens
* Added `<ProfilePreview>` component
* Componentised parts of the profile so that they can be reused in `<ProfilePreview>`. 
* Header is now fixed, footer sticks to bottom. 
* Introduced `.container` class in CSS, it is the layout element that contains everything else, it stays in the middle of the window, and is reused across all pages.
* We use new Protocol colors, removed old ones
* `<Modal>` component is now even more reusable, and used for display the profile preview on small screens

Still do to: 

* [x] ~~Find a good way to make `ProfilePreview` sometimes modal (on mobile), sometimes not~~ this is now done by using different components based on state
* [x] ~~Find a good way to make `ProfilePreview` stick, for some reason its `position: sticky` does not work~~ this is now done by making sure space underneath preview is taken up by a grid item (`::after` on the grid itself), so we now use `sticky` not `fixed`, so that it fits in the grid and overflow content can be accessed by scrolling
* [x] ~~Improve how org chart items are highlighted; basically we want them to be blue from left to right, even if they are indented.~~ made `<a>` full width, with left padding and icon position both based on `--nodeLevel` CSS variable 
* [x] ~~Maybe update URL with profile ID when a profile is previewed, so that people can copy paste it and it still works~~ this now works; it will do a doomed-to-fail look up with a hash that is not an ID, I think that is acceptable?

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -5,11 +5,15 @@ Vue.use(Vuex);
 
 export default new Vuex.Store({
   state: {
-    previewUserId: null,
+    profilePreview: {
+      userId: null,
+      lastTrigger: null,
+    }
   },
   mutations: {
     updatePreviewProfileId(state, update) {
-      state.previewUserId = update.newId; // eslint-disable-line no-param-reassign
+      state.profilePreview.userId = update.newId; // eslint-disable-line no-param-reassign
+      state.profilePreview.lastTrigger = update.trigger; // eslint-disable-line no-param-reassign
     },
   },
   actions: {
--- a/src/store.js
+++ b/src/store.js
@@ -9,7 +9,7 @@ export default new Vuex.Store({
   },
   mutations: {
     updatePreviewProfileId(state, update) {
-        state.previewUserId = update.newId;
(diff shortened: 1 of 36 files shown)


src/store.js
```diff
@@ -0,0 +1,18 @@
...
+
+Vue.use(Vuex);
+
+export default new Vuex.Store({
+  state: {
+    previewUserId: null,
+  },
+  mutations: {
+    updatePreviewProfileId(state, update) {
+      state.previewUserId = update.newId; // eslint-disable-line no-param-reassign
```
hidde: @fiji-flo our eslint does not like me changing something that's passed as param. Vuex docs [recommend](https://vuex.vuejs.org/guide/mutations.html#mutations) it. What are your thoughts?

PR #35, opened 2018-10-10
Add support for base url + templates.

fiji-flo:


akatsoulas:
@fiji-flo shall we close this?

PR #38, opened 2018-10-11
Remove non-functional PersonList on Home page

MichaelKohler:


fiji-flo:
Thanks. @hidde I think we can add this again (nothing gets lost here) once we're ready. But for now it removes clutter.

hidde:
Thanks for removing this, agreed we can remove for now. 

In the ultimate version it will be a listing _similar_ to this, we can refer to git history if we need to 👍

PR #39, opened 2018-10-11
Add support for base path (beta).

fiji-flo:
This is a bigger change than intended.
We also changed the state management for profile previews in the
orgchart. We now use the url params to trigger a profile preview.

Changed all <a> to <router-link>

### DIFF
--- a/src/components/OrgNode.vue
+++ b/src/components/OrgNode.vue
@@ -1,9 +1,9 @@
 <template>
-  <li :id="data.user_id" :class="'org-node' + ( data.user_id === $store.state.profilePreview.userId ? ' org-node--current' : '')">
-    <a :href="`/profile/${data.user_id}`" @click.prevent="previewProfile(data.user_id)" ref="openProfileTrigger">
+  <li :id="data.user_id" :class="'org-node' + ( data.user_id === this.$route.params.userId ? ' org-node--current' : '')">
+    <router-link :to="{ name: 'Orgchart', params: { userId: data.user_id } }">
       <span class="org-node__name">{{ data.first_name }} {{ data.last_name }}</span>
       <span class="org-node__title">{{ data.title }}</span>
-    </a>
+    </router-link>
     <ShowMore v-if="children.length > 0" :buttonText="`Expand ${data.first_name} ${data.last_name}`" :alternateButtonText="`Collapse ${data.first_name} ${data.last_name}`" :expanded="false" buttonClass="org-node__toggle" :transition="true">
       <template slot="overflow">
         <ul v-for="(child, index) in children" :key="index">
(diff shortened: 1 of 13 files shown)


src/components/OrgNode.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-  <li :id="data.user_id" :class="'org-node' + ( data.user_id === $store.state.profilePreview.userId ? ' org-node--current' : '')">
-    <a :href="`/profile/${data.user_id}`" @click.prevent="previewProfile(data.user_id)" ref="openProfileTrigger">
+  <li :id="data.user_id" :class="'org-node' + ( data.user_id === this.$route.params.userId ? ' org-node--current' : '')">
+    <router-link :to="{ name: 'Orgchart', params: { userId: data.user_id } }">
```
hidde: Looks like we're losing the `ref` here; it is used to store which link opened the org chart preview, so that we can return focus to it when the preview is closed. This may be as simple ad adding the `ref` to `router-link`?

src/router.js
```diff
@@ -21,7 +22,7 @@ const router = new Router({
       props: true,
     },
     {
-      path: '/org',
+      path: '/org/:userId?',
```
hidde: What are your thoughts on making this path `org-chart` instead of `org` so that it matches page's name?

src/views/PageOrgchart.vue
```diff
@@ -65,14 +65,13 @@ export default {
   },
   computed: {
     previewUserId() {
-      return this.$store.state.profilePreview.userId;
+      return this.$route.params.userId;
```
hidde: Ooh, I like it! 🎉

src/views/PageOrgchart.vue
```diff
@@ -94,9 +93,8 @@ export default {
       this.loading = false;
     },
     honourUriFragments() {
-      if (window.location.hash) {
-        this.$store.commit('updatePreviewProfileId', {
-          newId: window.location.hash.substr(1),
+      if (this.$route.params.userId) {
```
hidde: I think we can get rid of `honourUriFragments`; as I understand it we navigate using a route, and the routes are already being honoured.

There's no need to send `document.body` as the trigger; the purpose for it was that focus would get returned to the trigger, but focus gets returned to the body by default.

src/router.js
fiji-flo: We can change this again at some point. I moved it from `orgchart` to `org` since it overlapped with the orgchart service. But since this changes links to use the name `'Orgchart'` internally it's easy to change this once we settled on a schema.

src/components/OrgNode.vue
fiji-flo: I solved it slightly different. How do you feel about it?

src/components/ProfilePreview.vue
```diff
@@ -77,19 +77,13 @@ export default {
...
-      if (elementToReturnFocusTo) {
-        elementToReturnFocusTo.focus();
-      }
+      this.$router.push({ name: 'Orgchart' });
+      this.lastActive.focus();
+      this.lastActive = document.body;
     },
   },
   mounted() {
+    this.lastActive = document.activeElement || document.body;
```
hidde: NIce . This makes sense. 

I think we don't need the `|| document.body` fallback, afaik `document.activeElement` always exists.

src/router.js
hidde: 👍

PR #86, opened 2018-10-23
Use helm for templates.

akatsoulas:
* Remove trailing spaces.


akatsoulas:
Small update so that we can use helm everywhere in our repos.

hidde:
@akatsoulas this PR currently conflicts in the Profile.vue file, it looks like you can remove the changes in that file from this PR as they are not Helm-related. I believe they have been fixed in master.

akatsoulas:
@hidde I updated the PR to drop the changes in Profile.vue

Issue #70, opened 2018-10-16
desktop user menu should not span vertical


rfirebaugh:
Currently desktop user menu spans the entire vertical. Instead it should just appear as designed in zeplin. See attached screenshot mock.

<img width="549" alt="screen shot 2018-10-16 at 7 44 20 am" src="https://user-images.githubusercontent.com/43189936/47024894-70b28200-d117-11e8-800c-ea8bc2ca1d70.png">

hidde:
(Please ignore the above commits; I tested if they would close this issue. They did not.)

PR #88, opened 2018-10-24
[PR] Make user menu match new designs, add icons

hidde:
Addresses #70 and DP-353

New things:
- `<Icon>` component (added icons both in assets folder and inline in the `<Icon>` component so that the properties we need on presentational icons are in one place (`aria-hidden`, `role=presentation` etc). This will need improvement later
- lots of tweaks to colors, spacing, layout, mobile version

Signed-off-by: Hidde de Vries <hidde@hiddedevries.nl>


hidde:
Hey Unitobot, that's not what I wanted. Reopening

PR #109, opened 2018-11-01
Lots of things

fiji-flo:
- use graphql data in user menu
- fix view in orgchart (scrolling and expanding to person)
- use graphql for access groups
- use structured data for ssh/pgp keys

We copy the `user-demo.png` in the Dockerfile since it's no longer considered an asset by Vue. We should fix that.

### DIFF
--- a/src/components/OrgNode.vue
+++ b/src/components/OrgNode.vue
@@ -5,10 +5,10 @@
       <span class="org-node__name">{{ data.first_name }} {{ data.last_name }}</span>
       <span class="org-node__title">{{ data.title }}</span>
     </router-link>
-    <ShowMore v-if="children.length > 0" :buttonText="`Expand ${data.first_name} ${data.last_name}`" :alternateButtonText="`Collapse ${data.first_name} ${data.last_name}`" :trace="trace" :prefix="prefix" buttonClass="org-node__toggle" :transition="false" :expanded="firstLevel">
+    <ShowMore v-if="children.length > 0" :buttonText="`Expand ${data.first_name} ${data.last_name}`" :alternateButtonText="`Collapse ${data.first_name} ${data.last_name}`" :trace="trace" :prefix="prefix" buttonClass="org-node__toggle" :transition="false">
       <template slot="overflow">
         <ul v-for="(child, index) in children" :key="index">
-          <OrgNode :children="child.children" :data="child.data" :id="index" :prefix="`${prefix}-${id}`" :trace="trace"></OrgNode>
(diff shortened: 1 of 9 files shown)


src/components/OrgNode.vue
```diff
@@ -44,11 +43,6 @@ export default {
       }
     }
   },
-  computed: {
```
hidde: I suspect something happened to replace this? Can't see what it is, but if there is a replacement, I'm good with that! 🎉

src/components/Profile.vue
```diff
@@ -177,13 +177,14 @@ export default {
     description: Object,
     created: Object,
     lastModified: Object,
-    pgpPublicKeys: Object,
-    sshPublicKeys: Object,
+    pgpPublicKeys: Array,
```
hidde: Aaah… that's what I missed when I tried to fix this last night 😄

src/components/TopBar.vue
```diff
@@ -48,80 +67,81 @@ export default {
...
-      .top-bar__user-menu-toggle img {
-        margin-right: 0;
-        border-radius: var(--imageRadius);
-      }
-      .top-bar__user-menu-toggle .show-more__button-text {
-        position: absolute;
-        left: -9999em;
-        top: -9999em;
-      }
+.top-bar__bar {
```
hidde: ok ok 😄 If you prefer no indentation, let's change it across the board. I can do that sometime in next few weeks.

src/components/functional/ShowMore.vue
```diff
@@ -49,7 +45,7 @@ export default {
   watch: {
     trace() {
       if (this.trace || this.prefix) {
-        this.expanded = this.prefix === this.trace || this.trace.startsWith(`${this.prefix}-`) || this.expanded;
+        this.expanded = this.trace.startsWith(`${this.prefix}-`) || this.expanded;
```
hidde: Ah, it moved to here! Note that we use `ShowMore` in all places where we show/hide something (I might rename to `ShowHide` or `Toggle` at some point).

src/components/TopBar.vue
fiji-flo: Sorry, for that.

If you have a strong feeling about that I can try to teach my editor to accept this :P

src/components/functional/ShowMore.vue
fiji-flo: Jupp, I'm aware. When I played around with this, everything started to pop out.

PR #114, opened 2018-11-08
Add Picture component with identicons.

fiji-flo:


hidde:
I would prefer us to do the identicon image generation on the back-end, for two reasons:

* user doesn't need to download the library
* user's CPU doesn't need to generate the identicom

If time's our constraint, I'm happy for this to be merged in.

src/components/Person.vue
```diff
@@ -34,6 +36,9 @@ export default {
     funTitle: Object,
     picture: Object,
   },
+  components: {
+    Picture,
```
hidde: Idea: can we name this `Avatar` or `UserPicture`? For me this name is too close to HTML's native `picture` element, and it has a more specific use then being a picture (ie it's only for pictures that are people's profile picture, not other pictures, if we'll have any in this profile-centric site :-) ).

src/components/Picture.vue
```diff
@@ -0,0 +1,56 @@
...
+      if (this.picture === null) {
+        sha256(this.username).then((hash) => {
+          const identicon = new Identicon(hash, { size: this.size, format: 'svg' });
+          this.src = `data:image/svg+xml;base64,${identicon.toString()}`;
+        });
+      }
+    },
+  },
+};
+</script>
```
hidde: great work!

fiji-flo:
I changed the name to `UserPicture`

I see you concern regarding performance. I looked at the implementation and it looks solid. However, you're right that we could / should move this to the back-end. But it wasn't straight forward to design. When I ended up at: oh I just do a fetch for the identicon if there's no image, I basically had to choose between network or CPU.

PR #115, opened 2018-11-12
Add DinoType (Staff/Contributor)

hidde:
Notable changes: 

* `size` in `UserPicture` is now a `Number`
* `size` in `UserPicture` is now also used to determine what the Staff/Contributor label ('DinoType') looks like
* styling of `person__photo` in various places is now part of `UserPicture`
* optional `dinoType` prop in `UserPicture` determines if and which label is shown; this is now set to 'Staff' everywhere, we'll want to set it to something more useful when we have real data
* we now have just one link in a `Person`
  * when you tab with keyboard, each person is one tab (was: two)
  * when using screenreader, each person gets read out once (was: twice)

### DIFF
--- a/src/components/UserMenu.vue
+++ b/src/components/UserMenu.vue
@@ -3,7 +3,7 @@
     <div class="user-menu__header">
       <button class="user-menu__close-avatar" type="button" @click="sendCloseEvent()">
         <span class="visually-hidden">Close user menu</span>
-        <UserPicture :picture="picture.value" :username="userId.value" size="40"></UserPicture>
+        <UserPicture :picture="picture.value" :username="userId.value" :size="40"></UserPicture>
       </button>
       <div class="user-menu__name">
         <span class="user-menu__header-name">{{ firstName.value }} {{ lastName.value }}</span>
@@ -149,11 +149,8 @@ export default {
         border: 1px dotted currentColor;
       }
       .user-menu__close-avatar {
-        margin-right: 1.5em;
+        margin-right: 1em;
       }
-        .user-menu__close-avatar img {
-          min-width: 2.75em;
-        }
       .user-menu__close-button {
         margin-left: auto;
       }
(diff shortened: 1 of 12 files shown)


src/components/UserMenu.vue
```diff
@@ -3,7 +3,7 @@
     <div class="user-menu__header">
       <button class="user-menu__close-avatar" type="button" @click="sendCloseEvent()">
         <span class="visually-hidden">Close user menu</span>
-        <UserPicture :picture="picture.value" :username="userId.value" size="40"></UserPicture>
+        <UserPicture :picture="picture.value" :username="userId.value" :size="40"></UserPicture>
```
fiji-flo: wouldn't `size=40` work here, too?

src/components/UserPicture.vue
```diff
@@ -27,30 +31,113 @@ function sha256(str) {
...
     updateUserPicture() {
-      if (this.picture === null) {
+      if (this.picture === null || this.picture === '/beta/img/user-demo.png') {
         sha256(this.username).then((hash) => {
           const identicon = new Identicon(hash, { size: this.size, format: 'svg' });
           this.src = `data:image/svg+xml;base64,${identicon.toString()}`;
         });
       }
     },
+    decidePictureCategory(size) {
```
fiji-flo:
```js
    decidePictureCategory() {
      if (this.size <= 40) {
        this.dinoTypeSize = 'small';
      } else if (size <= 100) {
        this.dinoTypeSize = 'medium';
      } else {
        this.dinoTypeSize = 'large';
      }
      this.modifier = `user-picture--${this.dinoTypeSize}`;
    }
```
would work, too.

src/components/UserMenu.vue
hidde: With `:size` we are able to pass it as a JS expression, so that it is sent down as a `Number`.

src/components/UserPicture.vue
hidde: Ooh nice, like!

PR #124, opened 2018-11-15
Make URLs and page titles more descriptive

hidde:
I guess this is a little opinionated… I think with our new `/api/v3/graphql` URLs we're safe to update the URLs to be a little better?


src/router.js
```diff
@@ -17,25 +17,25 @@ const router = new Router({
       component: PageHome,
     },
     {
-      path: '/p/:userId',
+      path: '/profile/:userId',
```
fiji-flo: I really liked the short path. Do you have string feelings about this?

PR #128, opened 2018-11-16
adding label for Calico namespaceSelector

danielhartnell:
Calico allows us to block traffic from outside a namespace but it will use a `namespaceSelector` which needs a label rather than a namespace name. I'll create similar pull requests for the other services. Moving forward, it would be useful to add a name label to any new namespace just in case we need to use it.


hidde:
Hi @danielhartnell, do we still need this one to merge in?

PR #175, opened 2018-12-19
Avoid modal overlay when highlighting user in org chart on mobile, allow manual opening

hidde:
Changes:

* org chart links now go to the regular `OrgChart` route instead of `OrgChartHighlight`; they leverage the same template. A different route name was crucial to making it work, as we now use this extra `openedFromOrgChart` parameter to determine whether or not to open a modal on mobile, and the param-change did not populate when routing from user to same user (specifically: when opening a highlighted user, it was a route to `OrgChartHighlight` with that username, which was the page we were already on)
* removed `vuex ` for now, we only used it to store desktop vs mobile, which we did not use anywhere else, I believe this alone does not warrant having users download an extra library.


hidde:
Thanks @fiji-flo ! 

I've updated two more things:

* we no longer hide the 'view in org chart button' on mobile
* more icons are now inline svgs so that they are visible in high contrast mode

PR #180, opened 2019-01-16
Make first profile card editable; use new endpoint for profile

hidde:
A summary of changes: 

* reorganised structure of first card in the profile; now has a component for editing the first card and one for viewing it and the `editMode` variable in `data` (candidate for better abstracting when more cards become editable) 
* `EditPersonalInfo` does a `mutation` query, the Apollo cache is updated when successful and profile view is automagically updated
* new `<Toast>` component, which displays a notification that the whole profile page. Works by updating its `content` prop. It is a live region of the ‘assertive’ type, so gets read out by screenreaders immediately. It autoresets with a `setTimeout`.

Known issues, things we'll likely revisit:

* due to using new endpoint (`/api/v3/graphqlMut`), some fields, like `languages`,`manager` and `directs` are now hardcoded to keep it working, plan is to replace them by real thing as soon as that goes exposed


hidde:
@fiji-flo @Gregoor I'm merging my edit work into a branch named `add-edit-mode`, as it is pointing to the new GraphQL endpoint / using the new schema. It feels too nightly to be in `master` for now.

PR #181, opened 2019-01-22
Add custom select and use for pronouns, field visibility settings

hidde:
### New components

* `Options` and `Option`, which together can be used as a custom `select`, under the hood they are a set of radio buttons, what your see is their `label`

### New mixin

* `DisplayLevelsMixin`: returns, given a field, what the settable display levels are for that field

### Prettier compatible

* To make merging easier, I've prettified all the things.

## Features

* Custom select can have a label and/or an icon. There are props to set whether either should be visible in collapsed and expanded state. So you could have any of the below: 
* when collapsed, only show icon, when expanded, show icon and label
  * always show icon and label
  * only show labels
  * It works with keyboard.
* Extendable to have different styling (for example zebra striping… or wildly different)

## Known issues

* ~In current version, multiple instances of the visibility selector will sometimes update the top one. I'll investigate where it leaks (steps taken: unique `ref`s; separate component instances)~
* `ContactMe` and `Options` share a lot of styles (but not (yet) markup). I should refactor this.

## To do 

* ~Make mapping between `value` of `Options` and the page that uses it (and needs to send it as part of a mutation)~ done, now supports `v-model`
* ~Add downwards pointing icon in button; add checkmark for selected~
* Add optional zebra stripes
* Make it automagically position in a place where there is space (on the left if button is on the right of viewport, on the right if button is on the left of the viewport; note to self: [Popper](https://github.com/FezVrasta/popper.js) specialises in this)


hidde:
@Gregoor would you be able to review this? I'd like to merge it into `add-edit-mode`

src/components/Profile.vue
```diff
@@ -317,6 +318,7 @@ export default {
     Person,
     ProfileNav,
     ReportingStructure,
+    Options,
```
Gregoor: Looks like this isn't used in this file

hidde:
Thanks @Gregoor. And you're right it was not used, have removed.

PR #182, opened 2019-02-04
[WIP] Profile Photo

Gregoor:
Cropping works (tested in FF & Chrome), consumable like a library with a simple API.

TODO:
- [x] integrate it
- [x] make the modal look more like it does in [zeplin](https://app.zeplin.io/project/5b61ec0e22302659883e6834/screen/5c359cd5252fd04baf6829c9). @hidde should if I make this heading style, and the desktop max-width the default modal style?
- [ ] make it close properly, depends on #186 
- [x] once it feels right maybe package it into its own library + put it on NPM


src/assets/js/cr-op/cropper-element.js
```diff
@@ -0,0 +1,130 @@
...
+    setStyles(box, CROPPER_BOX_STYLE);
+    if (i > 3) {
+      setStyles(box, { 'border-top': 0 });
+    }
+    if (i % 3 !== 0) {
+      setStyles(box, { 'border-right': 0 });
+    }
+
+    if (i !== 5) {
+      const handle = document.createElement('div');
```
hidde: should this be a `button`?

Gregoor: Good question. I'm not sure. It's draggable, simply clicking it doesn't do anything. Is that still considered a button?

This also bring up the good point that this can not be used with the keyboard yet.

hidde: I _think_ it's more a `button` than it is a `div`, but there's no (or little) end user benefit to it if it doesn't also work with keyboard. Would it be doable to connect arrow up / right / down /  left to whatever dragging in that direction does?

Gregoor: Doable yes, though I think it'd take some time to get the UX right. I see two approaches right now:
1. what you said: every handle is focusable and the arrows keys do what dragging in that direction would've done
2. the whole thing is focusable, arrow keys move it, holding shift changes size in that direction

I think 2) is preferable, but I guess there'd be no way to communicate that interaction with accessibility attributes alone, would it?

hidde: Agreed 2 is preferable. Maybe we can show a toast that explains to use shift if we see people use their keyboard?

Gregoor:
@hidde I'll squash and merge this in, alright? Keyboard shortcuts aren't part of this yet, but I'll make sure we track it in JIRA.

PR #183, opened 2019-02-08
Prettier

Gregoor:
### What's in it?
Stop using ESLint for stylistic rules and start using Prettier instead. Also prettify all the existing files.

### Pros
- build log less cluttered
- no need to think about code styling while coding, prettier runs before every commit (only on changed files)
- overall: better DX

### Cons
- basically every vue file & line was touched, so if this is merged `git blame` will be useless for a while
- there was semantic indentation in the vue styles which this destroyes. One possible remedy would be to use [CSS Nesting](https://github.com/jonathantneal/postcss-nesting) which is supported by [PostCSS Preset Env](https://github.com/csstools/postcss-preset-env), imo a nicer solution overall. We're already using Babel Preset Env which is the same thing for JS, i.e. transpile to the browsers that match the browserslist query.

LMK if you have any concerns or questions


hidde:
Ok let's do this! thanks for also updating all files to match what Prettier wants.

Btw CSS Nesting looks nice, but I think it may make things too complex for us. The nesting I did there was to mimick the structure in the DOM, so that it is more clear what's a child of what. The nesting that CSS Nesting offers seems to bring some of Sass's functionality (a now being standardised flavour of it, TIL), which is powerful, but makes it less trivial to see which style rules we apply to what. Unindented CSS is fine for now.

Gregoor:
Fair enough! Thanks for taking the time to review!

PR #188, opened 2019-02-28
Lift options state up

Gregoor:
- Rename `Options` to `Select`
- Lift selected value state up
  - Options maintains less state now, selected value is a prop, currently selected option computed based on it (and options)
  - @hidde a good read on that topic in the React context: https://reactjs.org/docs/lifting-state-up.html Haven't found a similar article with a Vue contex
- turn DisplayLevelsMixin into a regular old-js file, as I needed the default display state
- removed defaultToFirst and made that the default behavior. So that it behaves more like regular-ol select's
- fill the EditPersonalInfo-s data based on all the given initialValues (less explicit, but also no repetition)

### DIFF
--- a/src/components/Select.vue
+++ b/src/components/Select.vue
@@ -33,7 +33,7 @@
           :icon="option.icon"
           :id="`option-${id}-${index}`"
           :bind="{ expandedShowIcon, expandedShowLabel }"
-          @input="$emit('input', value)"
+          @input="$emit('input', $event)"
           @close-list="closeList"
         />
       </ul>
@@ -46,16 +46,12 @@ import Icon from '@/components/Icon.vue';
 import Option from '@/components/Option.vue';
 
 export default {
-  name: 'Options',
+  name: 'Select',
   props: {
     label: String,
     id: String,
     value: String,
     options: Array,
-    defaultToFirst: {
-      type: Boolean,
-      default: false,
-    },
     collapsedShowIcon: {
       type: Boolean,
       default: true,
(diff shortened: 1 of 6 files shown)


src/components/Select.vue
```diff
@@ -81,9 +78,9 @@ export default {
       if (this.open) {
         this.open = false;
       } else {
+        const list = this.$refs[`optionList-${this.id}`];
         const optionToFocus =
-          this.$refs[`optionList-${this.id}`].querySelector('input:checked') ||
-          this.$refs[`optionList-${this.id}`].querySelector('input');
+          list.querySelector('input:checked') || list.querySelector('input');
```
hidde: Thanks, readability++

src/components/Select.vue
```diff
@@ -45,16 +46,12 @@ import Icon from '@/components/Icon.vue';
...
 
 export default {
-  name: 'Options',
+  name: 'Select',
   props: {
     label: String,
     id: String,
     value: String,
     options: Array,
-    defaultToFirst: {
```
hidde: Excellent idea to just make it the default!

src/components/forms/EditPersonalInfo.vue
```diff
@@ -251,40 +250,17 @@ export default {
...
-      },
-      timezone: {
-        value: this.initialValues.timezone.value,
-        display: this.initialValues.timezone.display,
-      },
-      description: {
-        value: this.initialValues.description.value,
-        display: this.initialValues.description.display,
-      },
+      ...Object.entries(this.initialValues).reduce(
```
hidde: 🥇 this is great

PR #189, opened 2019-03-04
Add combobox and use for pronouns, location

hidde:
This wraps the [combobo](https://github.com/dequelabs/combobo) library to create a combobox component

### DIFF
--- a/src/components/Combobox.vue
+++ b/src/components/Combobox.vue
@@ -56,9 +56,9 @@ export default {
       });
     },
     checkInput(e) {
-      const alwaysAllowed = [9, 13, 27, 16, 8]; // ignore tab, enter, escape, shift, backspace
+      const alwaysAllowed = ['Tab', 'Enter', 'Escape', 'Shift', 'Backspace'];
 
-      if (alwaysAllowed.indexOf(e.which) >= 0) {
+      if (alwaysAllowed.includes(e.key)) {
         return;
       }
 
--- a/src/components/Combobox.vue
+++ b/src/components/Combobox.vue
@@ -25,12 +25,11 @@
 import Combobo from 'combobo';
 
 export default {
-  name: 'Autocomplete',
+  name: 'Combobox',
   props: {
     id: String,
     source: Array,
     value: String,
-    initialValue: String,
     allowCustomInput: {
       type: Boolean,
       default: true,
--- a/src/components/Combobox.vue
+++ b/src/components/Combobox.vue
@@ -6,6 +6,7 @@
       class="combobox__input"
       ref="comboboxInput"
       :value="value"
+      @keydown="checkInput"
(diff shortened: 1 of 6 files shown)


src/components/Combobox.vue
```diff
@@ -0,0 +1,115 @@
...
+      this.combobo.addEventListener('selection', () => {
+        this.$nextTick(() => {
+          this.$emit('input', this.$refs.comboboxInput.value);
+        });
+      });
+    },
+    checkInput(e) {
+      const alwaysAllowed = [9, 13, 27, 16, 8]; // ignore tab, enter, escape, shift, backspace
+
+      if (alwaysAllowed.indexOf(e.which) >= 0) {
```
Gregoor: [`array.includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) would be more explicit.

Also `indexOf` return `0` if the first element matches, so checking for `>= 0` might not be what you want here.

src/components/Combobox.vue
```diff
@@ -0,0 +1,115 @@
...
+    source: Array,
+    value: String,
+    allowCustomInput: {
+      type: Boolean,
+      default: true,
+    },
+    placeholder: String,
+  },
+  methods: {
+    init() {
```
Gregoor: I'd personally just move that to `mounted()` to reduce indirection.

src/components/Combobox.vue
```diff
@@ -0,0 +1,115 @@
...
+
+      this.combobo = new Combobo({
+        input: comboboxInput,
+        list: comboboxList,
+        options: this.optionEls,
+        activeClass: 'combobox__option--selected',
+      });
+
+      this.combobo.addEventListener('selection', () => {
+        this.$nextTick(() => {
```
Gregoor: Why `$nextTick`?

src/components/Combobox.vue
hidde: Hm, `array.includes` reads better, but does not work in IE11, not sure if we want to polyfill, I'll double check our browser requirements.

src/components/Combobox.vue
hidde: Haha, I was probably prematurely optimising, not sure if we'll meet a case where we'd want to init on a different lifecycle hook. Have moved to `mounted()` now.

src/components/Combobox.vue
hidde: For some reason, I was unable to emit the input event on `selection`, waiting for the next tick (DOM update) fixed that.

src/components/Combobox.vue
hidde: Doublechecked, IE11 is not a requirement, `array.includes` it is!

PR #191, opened 2019-03-06
Folderise and add edit route

hidde:
* Components are now organised with the things they depend on, which is sort of like by page
* Edit is now its own route, it takes a `section` query parameter which specifies the section to be edited


hidde:
Closing this, will separate to reorganising folders (#192 ) and creating edit route

PR #195, opened 2019-03-11
Make edit a route

hidde:
* Editing now has its own route. It takes a parameter in which we specify which card is editable (it's always one at this point)
* New `EditMutationWrapper` component contains logic to do mutation, Save/Cancel buttons, the surrounding form, so that we don't need to repeat that code across cards
* New `EmptyCard` component is used for empty cards (for if user doesn't have content for that particular card)

Note that View Profile and Edit Profile are still in the same component, the Profile component. but depending on the route parameter one or more cards can be in edit mode.


src/components/profile/Profile.vue
```diff
@@ -2,31 +2,14 @@
   <main class="profile container">
     <div
       :class="
-        'profile__section' + (this.editMode ? ' profile__section--editing' : '')
+        'profile__section' +
+          (this.editing && this.editing === 'personal-info'
```
Gregoor: could be simplified to `this.editing === 'personal-info'`

src/components/profile/Profile.vue
```diff
@@ -2,31 +2,14 @@
...
-          funTitle,
-          picture,
-          location,
-          description,
-        }"
-        @toggle-edit-mode="toggleEditMode"
-      />
       <EditPersonalInfo
-        v-else
+        v-if="this.editing && this.editing === 'personal-info'"
```
Gregoor: same as above

src/components/profile/edit/EditMutationWrapper.vue
```diff
@@ -0,0 +1,218 @@
...
+
+export default {
+  name: 'EditMutationWrapper',
+  props: {
+    initialValues: Object,
+    formName: String,
+    editVariables: Object,
+  },
+  methods: {
+    cancelEdit() {
```
Gregoor: looks like this isn't called?

src/components/profile/Profile.vue
hidde: Ah you are right, fixed!

src/components/profile/edit/EditMutationWrapper.vue
hidde: Good point, it's now a route, removed.

hidde:
Thanks @Gregoor. I'll merge it in sometime today, we can (and probably should) iterate from here.

PR #207, opened 2019-04-02
Fix broken icon block in safari.

fiji-flo:


src/components/ui/IconBlockList.vue
```diff
@@ -15,7 +15,7 @@ export default {
 
 <style>
 .icon-block-list {
-  padding-left: 1em;
+  padding-left: 0em;
```
jens1o: `padding-left: 0;`?

PR #208, opened 2019-04-02
full block edit loader

fiji-flo:


src/components/profile/edit/EditMutationWrapper.vue
```diff
@@ -120,6 +125,18 @@ export default {
 </script>
 
 <style>
+.mutation-form + .loading {
+  position: absolute;
+  z-index: 1;
```
hidde: Ideally this would use the vars too, but I'll take care of that

PR #212, opened 2019-04-03
Improve dropdown (fixes DP-1076, DP-1137, DP-1142, DP-1143)

hidde:
* Make username privacy settings non-editable (DP-1142)
* Make a guess about where the custom select should show (left, right, center), kept this fairly simple to cover most use cases (DP-1076)

### DIFF
--- a/src/components/ui/Select.vue
+++ b/src/components/ui/Select.vue
@@ -1,5 +1,8 @@
 <template>
-  <div :class="'options' + (position ? ' ' + `options--${position}` : '')">
+  <div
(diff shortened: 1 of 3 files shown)


src/components/ui/Select.vue
```diff
@@ -1,5 +1,8 @@
 <template>
-  <div class="options">
+  <div
+    :class="'options' + (position ? ' ' + `options--${position}` : '')"
```
fiji-flo: that's the same as ```(position && ` options--${position}`)```

hidde: Good point! Fixed

PR #213, opened 2019-04-03
enable markdown in description

fiji-flo:


hidde:
I assume we do, but just to double check, do we sanitise the profile? Afaik HTML is also valid Markdown

fiji-flo:
Good thought. I read about it and the marked sanitizer is disabled by default and apparently not the best. Added [insane](https://github.com/bevacqua/insane) looks like it's doing good job. It catches things like ``.

PR #231, opened 2019-04-05
Fix display of cards that have no data

hidde:
This PR: changes deciding which sections to show, adds edit button to empty card states if user is on own profile.

Other changes:

* uses `EditButton` component instead of repeating the same; it's also a `RouterLink` now

### DIFF
--- a/src/components/profile/Profile.vue
+++ b/src/components/profile/Profile.vue
@@ -97,9 +97,16 @@
       ></EditAccounts>
       <ViewAccounts v-else v-bind="{ uris, userOnOwnProfile }"></ViewAccounts>
     </section>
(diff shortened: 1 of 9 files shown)


src/components/profile/Profile.vue
```diff
@@ -242,9 +265,9 @@ export default {
       return {
         relations: this.staffInformation.staff,
         contact: true,
-        accounts: this.uris.values !== undefined,
-        languages: this.languages.values !== undefined,
-        tags: this.tags.values !== undefined,
+        accounts: this.uris.values !== null,
+        languages: this.languages.values !== null,
+        tags: this.tags.values !== null,
```
fiji-flo: see our slack conversation

PR #248, opened 2019-04-08
support deleting avatars

fiji-flo:
- delete avatar will close the modal and use the identicon
- also change avatar canvas size to 264

### DIFF
--- a/src/components/ui/UserPicture.vue
+++ b/src/components/ui/UserPicture.vue
@@ -13,27 +13,8 @@
 </template>
 
 <script>
-import Identicon from 'identicon.js';
-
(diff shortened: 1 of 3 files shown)


src/components/ui/UserPicture.vue
```diff
@@ -13,28 +13,9 @@
 </template>
 
 <script>
-import Identicon from 'identicon.js';
-
+import generateIdenticon from '@/assets/js/identicon-avatar';
```
hidde: `master` doesn't seem to have this file yet, should be part of this PR?

fiji-flo: :+1:

PR #249, opened 2019-04-08
catch error on invalid timezone

fiji-flo:


public/error.html
```diff
@@ -9,10 +9,7 @@
       href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Zilla+Slab"
       rel="stylesheet"
     />
-    <link
-      href="https://dinopark.k8s.dev.sso.allizom.org/beta/css/app.css"
-      rel="stylesheet"
-    />
+    <link href="/beta/css/app.css" rel="stylesheet" />
```
hidde: good find, I added full path to test and forgot to revert back to `/beta/css/app.css`

PR #255, opened 2019-04-09
support image orientation

fiji-flo:
This slows downs things, because we need to parse the image before loading it. However we can test and revert if it's to slow.


hidde:
Hm, it also [makes the bundle bigger](https://bundlephobia.com/result?p=blueimp-load-image@2.21.0). Should we think about creating an image edit bundle that we only load when someone edits their personal info? (So that people who just browse profiles don't download it?)

Approving for now, change looks good to me!

PR #264, opened 2019-04-09
move to /

fiji-flo:


hidde:
Woa, we still had quite a lot hardcoded. Was going to remind you of the search thingy, but see you've even taken care of that 🙏 

Yay, this is a great step!

PR #271, opened 2019-04-12
Replace the term "contributor" with "volunteer"

openjck:
Mitchell said something about these terms many years ago which always
stuck with me. Paraphrased, she said something like this:

"We are all contributors. Some of us are paid. Others volunteer their
time."

Although I can't find any formal guidelines about this, I notice that
Mozilla generally follows Mitchell's guidance. For example, the "Our
Contributors" page lists both paid and unpaid people,[1] we refer to
unpaid work as "volunteer opportunities" on mozilla.org,[2] and the
careers page uses the phrase "paid and volunteer contributors."[3]

[1] https://www.mozilla.org/credits/
[2] https://www.mozilla.org/contribute/
[3] https://careers.mozilla.org/

### DIFF
--- a/src/components/search/SearchScope.vue
+++ b/src/components/search/SearchScope.vue
@@ -35,13 +35,13 @@
         />
         <label for="search-who-staff">Staff</label>
         <input
-          id="search-who-contributors"
+          id="search-who-volunteers"
           type="radio"
-          value="contributors"
+          value="volunteers"
           name="who"
           v-model="who"
         />
-        <label for="search-who-contributors">Contributors</label>
+        <label for="search-who-volunteers">Volunteers</label>
       </div>
     </fieldset>
     <input


openjck:
This corresponds to the following pull request in dino-park-search:

https://github.com/mozilla-iam/dino-park-search/pull/47

fiji-flo:
Hey @openjck thanks for looking at this. I will start a conversation with our UX designer and we would change it everywhere. I think you made a good case there.
In terms of you're PR. This would break the back-end. You're changing the `value` of and `input` which ends up in the search request. I can change the back-end to support both for now. Then we can merge this after we got a go from Product/UX.

fiji-flo:
Hhaha, I should look at all notifications before replying. Ignore my back-end comment you took already care of that.

Just waiting for Product/UX then.

jens1o:
What's the current status here?

openjck:
I'm happy to update the branch. I'm blocked on a product decision.

PR #272, opened 2019-04-16
Add Tooltip and reusable Popover component

hidde:
This adds two new components: `Tooltip` and `Popover`.

`Popover` is now reused for its styling and positioning in: `ContactMe`, `Select` (this includes privacy settings) and `Tooltip`.

Also added:

* button text in `ShowMore`  can now be ‘visually hidden’ with a prop. We already hacked this in the top bar with button text for user menu and search. Hack is now removed in those places, the new prop is used instead. It is also used for the tooltip opener.


hidde:
@fiji-flo I've addressed your comments:

* ContactMe falling underneath the profile nav on desktop: the `Popover` now takes a `layer` prop to override the default, and we use it for `ContactMe`
* the tooltip size on small screens was weird, makes more sense now (90% of the viewport, with a variable maximum that we also use to decide about how much available space there is)

PR #274, opened 2019-04-17
Add expand/collapse all toggles to org chart

hidde:
This combines the All|Contributors|Staff toggle and what we required for the org chart into one reusable `Toggle` component.

This component lets users toggle between multiple options, an option can have text and/or an icon.

In the org chart, it's icons (proper expand/collapse icons to be added):

Three buttons with icons

In the search results, it's text:

toggle with all, staff and contributors options

In theory one could add both an icon and a text here, the component has the CSS to support both (this is a fictional example for demonstration purposes): 

example with options that have icons and text

also cc'ing @mbransn as an fyi, let's iterate in design QA if required


fiji-flo:
part of #314

PR #278, opened 2019-04-23
Fix click outside without breaking org chart

hidde:
In a previous version, my clicked outside broke the org chart, this no longer does that.

It is based on the tooltip branch, so we best merge this after that.


hidde:
Closing as this is now fixed in the tooltip branch as that had some of the problems and needed these fixes.

PR #282, opened 2019-04-23
ShowMore: move stop propagation

fiji-flo:
- this fixes an issue with show mores not closing when
  opening another


hidde:
Nice, thanks!

PR #285, opened 2019-04-24
Increase avatar picture quality in user menu

hmitsch:
**Benefit:** 
The user avatar is less blurry.

**Drawback:**
This change requires browsers to download more data because the image will be delivered in a higher quality.

**Mitigation:**
The avatar quality is only increased for the user menu, **not** for the org chart or profile page. This limits the bandwidth increase to 1 avatar picture per user.

**Before:**
image

**After:**
image

### DIFF
--- a/src/components/ui/TopBar.vue
+++ b/src/components/ui/TopBar.vue
@@ -72,7 +72,7 @@
             <UserPicture
               :picture="user.picture.value"
               :username="user.primaryUsername.value"
-              :size="264"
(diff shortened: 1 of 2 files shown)


hmitsch:
Adapted image size to **medium** following a quick conversation with @fiji-flo.

_Note to self:_ 
Supported image sizes are listed here: https://github.com/fiji-flo/dino-park-fossil/blob/master/src/send/resize.rs

PR #316, opened 2019-05-06
fix avatar editing

fiji-flo:


src/components/profile/edit/EditPictureModal.vue
```diff
@@ -110,7 +115,7 @@ export default {
       );
     },
     selectCrop() {
-      if (this.$refs.crop && !this.imgSrc) {
```
fiji-flo: this broke saving a new avatar :/

PR #322, opened 2019-05-07
Add a more button for > 100 search results.

fiji-flo:


src/pages/PageSearchResult.vue
```diff
@@ -60,13 +60,21 @@
...
         <strong>{{ this.$route.query.query }}</strong>
       </p>
-      <SearchResultList :results="results"></SearchResultList>
+      <SearchResultList :dinos="results.dinos"></SearchResultList>
+      <LoadingSpinner v-if="loading"></LoadingSpinner>
+      <Button
+        v-else-if="results.dinos.length < results.total"
+        class="search-results__button-more"
+        @click="fetchMore"
+        >more</Button
```
hidde: maybe make this Load More to match capitalisation/content in other action buttons?

src/pages/PageSearchResult.vue
```diff
@@ -87,37 +96,54 @@ export default {
...
+      const results = await this.fetchData();
+      this.results = results;
+      this.loading = false;
+    },
+    async fetchMore() {
+      this.loading = true;
+      const results = await this.fetchData(true);
+      this.results.dinos.push(...results.dinos);
+      this.results.next = results.next;
+      this.loading = false;
```
hidde: Quick idea: maybe at this point we could convey that an update happened through a `Toast`?

fiji-flo: There's a loading spinner when loading. I think an additional toast would be overkill here.

PR #323, opened 2019-05-07
Add language editing

hidde:
Updates compared to previous versions of add languages:

* Uses new interaction pattern as discussed with @mbransn 
* Like the latest version of `EditAccounts`, it now just uses one `languages` object for all the things
* Restores edit button in empty card


src/components/profile/edit/EditLanguages.vue
```diff
@@ -53,56 +59,78 @@ export default {
...
+      if (this.newLanguage.length > 0) {
+        this.addLanguage(this.newLanguage);
+        this.addingLanguage = false;
+      } else {
+        this.$refs.inputLanguage.focus();
+      }
+    },
     addLanguage(language) {
-      const id = Object.keys(this.languages).length + 1;
+      const id = this.languages.values[this.languages.values.length] + 1;
```
hidde: @fiji-flo Could you please help me with this? 🙏 The object needs to have a number for `k` that is higher than the previous one. I came up with this, but it breaks when user deletes a language, as then this number will be too low resulting in accidentally overwriting an existing entry.

Or do perhaps know of a smarter way to ensure the `k`'s in `language` are just always 1, 2, 3, 4… ?

Gregoor: `const maxKey = Object.keys(this.languages.value).reduce((max, key) => Number(key) > max ? Number(key) : max, 0)`

hidde: thanks so much @Gregoor ! (for reference: we decide as long as new keys are higher than any existing key, that's fine)

hidde:
Fixed merge conflict. This is now ready for review.

PR #334, opened 2019-05-14
Org-Chart Refactor & Fixes (DP-835)

Gregoor:
Might've gone a lil crazy on this one:

- moved org chart expanded/collapsed state into PageOrgChart
  - ➕  no more state syncing
  -  ➖ implicit dependency, OrgChart components need to be nested inside PageOrgChart
- built a simple subscription mechanism
  - ➕  only affected components get notified about state update
  -  ➖ a bit contraire to the whole reactive model

Unrelated change, but since I was already there:
- set the `--nodeLevel` custom property dynamically

### DIFF
--- a/src/pages/PageOrgchart.vue
+++ b/src/pages/PageOrgchart.vue
@@ -3,7 +3,7 @@
     <div class="org-chart-buttons" v-if="!loading">
       <button
         type="button"
-        @click="expand"
+        @click="expandAll"
         class="org-chart-buttons__row-left org-chart-buttons__control button--icon-only button"
         title="Expand all"
       >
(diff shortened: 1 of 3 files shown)


src/pages/PageOrgchart.vue
```diff
@@ -107,7 +106,35 @@ export default {
...
       desktopView: false,
-      baseState: 'normal',
+
+      expanded: {},
+      collapsed: null,
+      subscribers: {},
+    };
+  },
+  provide() {
+    const self = this;
```
fiji-flo: Maybe we can call this `orgChartPage` I love the nostalgia here, but my first though was why do we need this.

Gregoor: Haha I felt nostalgic writing it! Fair enough, I'll change it

Gregoor: Unfortunately we need it because Vue binds functions given through `provide` to the component they're being `injected` into

PR #338, opened 2019-05-17
org chart remember expanded state in store

Gregoor:
https://jira.mozilla.com/browse/DP-835

### DIFF
--- a/src/pages/PageOrgchart.vue
+++ b/src/pages/PageOrgchart.vue
@@ -96,6 +96,7 @@ export default {
     Toggle,
   },
   data() {
+    const { org } = this.$store.state;
     return {
       loading: false,
       post: null,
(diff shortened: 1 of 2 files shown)


src/pages/PageOrgchart.vue
```diff
@@ -196,7 +197,7 @@ export default {
         const orgchart = await data.json();
         this.tree = orgchart.forrest;
         this.loose = orgchart.loose;
-        if (this.$route.name !== 'OrgchartHighlight') {
+        if (!this.orgFromStore && this.$route.name !== 'OrgchartHighlight') {
```
fiji-flo:
```suggestion
        if (this.$route.name === 'OrgchartHighlight') {
          this.collapseAll();
        } else if (!this.orgFromStore) {
```
I would do collapse first when highlighting

PR #339, opened 2019-05-17
Make contact me less wide (DP-1267)

hidde:
I've made `maxWidth` for the popover a prop so that we can change it based on where it is used.

This sets a smaller `maxWidth` for the `ContactMe` popover, as we need less space now that we no longer display user names.


hidde:
@Gregoor oh, you're totally right, thanks! I believe overwriting props is not impossible in Vue, but it yields warnings and is certainly bad practice. Have updated it according to your recommendation

PR #342, opened 2019-05-17
auto-close user menu when links are clicked while still following links

Gregoor:
@fiji-flo I'm not happy with this just yet, I'm still using a timeout as links aren't followed for some reason when the menu closes immediately. The click/touch event is listenning in the bubble phase which should mean that the menu is closed after the link was opened, shouldn't it? I'm a little puzzled by this.
The timeout does the job but it's a dirty hack and who knows how it behaves on slow devices :/

@hidde there was some anti-patterning going on here, where the `TopBar` was setting state for its child `ShowMore` components. Data should only flow through props from Parent to Child. In this case I didn't find out the exact reasoning behind it, as `ShowMore` controlled its own state in all instances, so maybe just a relic of the past. Please LMK if I missed sth here.

### DIFF
--- a/src/components/_functional/ShowMore.vue
+++ b/src/components/_functional/ShowMore.vue
@@ -60,7 +60,6 @@ export default {
     alternateButtonText: String,
     buttonClass: String,
     transition: Boolean,
-    expanded: Boolean,
     closeWhenClickedOutside: {
       type: Boolean,
       default: false,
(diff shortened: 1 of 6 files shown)


hidde:
@Gregoor That was originally meant to allow for setting whether it should initially appear open or closed (you are completely right, I should have used a prop, like `isInitiallyOpen`). But… we don't have any `ShowMore`s where we want them to be initially open, so I am totally on board with your change to just set it to `false`.

PR #343, opened 2019-05-20
Make keys editable (DP-1060)

hidde:
* Keys now look like designs, are cut off visually after the first line
* Keys now have “Copy” button to copy to clipboard
* In edit view, each type of key has a heading + privacy control

Q: I [used `try`/`catch` but removed](https://github.com/mozilla-iam/dino-park-front-end/pull/343/commits/58a6d0e06ebc11049df55b2c14af7b40668fb7bf), because it doesn't actually seem to work, if I change 'copy' into 'banana' it doesn't throw an error. Should we check success differently or just always assume success since [`execCommand` with `copy` is well supported](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)?


Gregoor:
Looking good 🚀 🙂

PR #348, opened 2019-05-22
Update colleagues section (DP-1245)

hidde:
This adds a toggle to  switch between grid and list views:

two buttons, one with grid icon one with list icon

It also makes the avatars as big as they should be, and hides direct in 'Show more' button if there's more than 10. 

list of avatars on a grid with show more button underneath

And it updates the typography to be not bold: 

heading manages in regular not bold type

(fixes https://github.com/mozilla-iam/dino-park-issues/issues/32)


hidde:
Note that persistency is not added to this yet, will add if needed.

hidde:
This now stores the setting in `vuex`.

And I've added a [fix](https://github.com/mozilla-iam/dino-park-front-end/pull/348/commits/fecefbd4fcd274bdb01741822a88af04b0290406) for profiles that have very long words in them that would break the grid (tldr `word-wrap: anywhere` on profile intro ).

PR #349, opened 2019-05-24
timeout faster

fiji-flo:


fiji-flo:
500ms was a little high. That should work.

Gregoor:
It wasn't happening consistenly on my machine with <300ms. My hack might bite us for this 🙈

PR #360, opened 2019-06-04
Identities

fiji-flo:


src/App.vue
```diff
@@ -51,6 +51,9 @@ export default {
   --white: #fff;
   --black: #000;
 
+  --true-green: #12bc00;
```
hidde: Hm… our other colors match the [naming scheme in Protocol](https://protocol.mozilla.org/fundamentals/color.html), which I found helpful for seeing at a glance how the colors relate to each other (gray-20 is lighter than gray-30, yay!) 

But these colors don't seem to exist in Protocol (cc @mbransn), so I guess we can leave this for now

PR #363, opened 2019-06-05
Speed 2

fiji-flo:


src/pages/PageOrgchart.vue
```diff
@@ -304,7 +431,10 @@ export default {
   width: 2.5em;
   height: 2.5em;
 }
-.org-chart-buttons__control:hover {
+.org-chart-buttons__control:hover:disabled {
+  cursor: not-allowed;
+}
+.org-chart-buttons__control:hover:enabled {
   background-color: var(--blue-60);
```
hidde: In https://github.com/mozilla-iam/dino-park-front-end/pull/362, I am making this gray (same gray as border), as per @mbransn's request. Maybe we can do it here?

PR #373, opened 2019-06-10
Improve colleagues mobile (DP-1245)

hidde:
* user pictures were shrinking because they were flex items; set flex to `none` to avoid, as we always want them to be square
* force name and job title to be on one line with … if needed
* force header controls (like list/grid view on colleagues header) to next line on small screens
* rewrite of 'Show More' for colleagues: now relies on toggling `overflow: hidden`/`max-height` on a container, this makes it work without us having to know how many avatars fit on one line (we use Grid Layout's `auto-fit` to make it ‘as many as possible’)

### DIFF
--- a/src/components/profile/ReportingStructure.vue
+++ b/src/components/profile/ReportingStructure.vue
@@ -172,7 +172,7 @@ export default {
 .reporting-structure__directs--grid button {
   position: absolute;
   bottom: 0;
-  width: calc(100% - 2.5em);
+  width: calc(100% - 3.5em);
 }
 @media (min-width: 57.5em) {
   .reporting-structure__directs--grid button {
--- a/src/components/profile/ReportingStructure.vue
+++ b/src/components/profile/ReportingStructure.vue
@@ -149,6 +149,9 @@ export default {
   font-size: 1em;
   color: var(--gray-50);
 }
+.reporting-structure__directs {
+  min-width: 0;
+}
 .reporting-structure__directs--grid {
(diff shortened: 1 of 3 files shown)


src/components/profile/ReportingStructure.vue
```diff
@@ -95,10 +96,24 @@ export default {
...
+      ) {
+        this.managesExpandable = true;
+      }
+    }, 700);
+  },
   data() {
     return {
       related: new Related(),
-      initiallyShown: 10,
+      managesExpanded: false,
```
fiji-flo: these break between profiles. Both are not reset when switching to another profile.

src/components/profile/ReportingStructure.vue
```diff
@@ -134,4 +149,39 @@ export default {
...
+  max-height: none;
+  overflow: visible;
+}
+.reporting-structure__directs--expandable {
+  margin-bottom: 2em;
+}
+.reporting-structure__directs--grid .reporting-structure__directs-list {
+  display: grid;
+  grid-template-columns: repeat(auto-fit, 3em);
+  justify-content: space-between;
```
fiji-flo: Is this intended in case of 3 directs this will spread them across the while space, wouldn't it?

hidde: You're right, it would and we don't want that; will remove this line, it defaults to `start`.

hidde:
Closing this now in favour of #376

PR #377, opened 2019-06-11
Remove banner, leaving the component in place.

fiji-flo:


hidde:
Maybe also remove `showBanner` from `data`?

PR #396, opened 2019-07-05
blank pictures, hide related buttons, chunk up

fiji-flo:
- blank pictures when switching profiles
- hide list/grid button when only 0/1 directs
- load deps async

### DIFF
--- a/src/components/profile/edit/EditPersonalInfo.vue
+++ b/src/components/profile/edit/EditPersonalInfo.vue
@@ -31,8 +31,10 @@
     <div class="edit-personal-info">
       <div class="edit-personal-info__picture">
         <UserPicture
-          :picture="pictureData.value"
-          :username="loggedInUser.primaryUsername.value"
+          :avatar="{
+            picture: pictureData.value,
+            username: loggedInUser.primaryUsername.value,
+          }"
           :size="264"
           :isStaff="true"
         ></UserPicture>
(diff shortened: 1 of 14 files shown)


src/components/profile/edit/EditPersonalInfo.vue
```diff
@@ -419,7 +420,7 @@ export default {
     initialValues: Object,
   },
   components: {
-    Combobox,
+    Combobox: () => import('@/components/ui/Combobox.vue'),
```
jens1o: why is this loaded lazily, although (at least afaic) it is loaded everytime?

src/components/profile/edit/EditPictureModal.vue
```diff
@@ -106,7 +104,7 @@ export default {
     staffInformation: Object,
   },
   components: {
-    Crop,
+    Crop: () => import('./Cropper.vue'),
```
jens1o: same here?

src/components/profile/edit/EditPersonalInfo.vue
fiji-flo: I'm starting to chunk up the dependencies. This produces multiple chunks in a `npm run build` (check it out). See https://vuejs.org/v2/guide/components-dynamic-async.html
The goal is to lazy load them at some point. So theses components are only loaded when a user edits their profile.

jens1o: Ah, I see. Weird semantic, though

PR #400, opened 2019-07-18
fix(dino-park): making all "feedback" links be the same

andrew-sunada:
Provide a global set of links and for consistency, and refactor links.

### DIFF
--- a/src/components/ui/Card.vue
+++ b/src/components/ui/Card.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="card">
+  <div class="card" v-on:click="onCardClicked">
     <slot></slot>
   </div>
 </template>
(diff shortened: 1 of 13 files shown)


andrew-sunada:
Jira stories:
- https://jira.mozilla.com/browse/DP-1411
- https://jira.mozilla.com/browse/DP-1416

src/pages/PageHome.vue
```diff
@@ -11,81 +11,28 @@
...
+          target="_blank"
+          rel="noopener noreferrer"
+          class="card__link"
+        >
+          <div class="card__icon">
+            <Icon v-bind:id="card.icon" :width="64" :height="64"></Icon>
+          </div>
+          <div class="card__content">
+            <h2>{{ card.headline }}</h2>
+            <p v-html="card.description"></p>
```
fiji-flo: We produce nested anchors here, that's not w3c conform. I'm really not sure it there's a better way than Hidde did it with the pseudo element.

src/components/ui/Card.vue
```diff
@@ -1,23 +1,59 @@
...
 <script>
 export default {
   name: 'Card',
+  methods: {
+    /**
+     * Determine if card either has a 'primary link' defined by '.card__link'
+     * or if the clicked item is itself an anchor tag. If it has either, redirect
+     * according to link requirements.
+     */
+    onCardClicked(e) {
```
fiji-flo: I really get this. We're trying to support different / multiple links in the cards. Moving this to JS breaks things like _long press on → open in new tab_ on mobile or _middle click to open in new tab_ on desktop.
I would vote for either talking to @mbransn about using one the earlier proposals (not having the card clickable or not having a link in the text and make the whole card a link) or accept the original pseudo element version.

What do you think?

PR #407, opened 2019-07-23
fix(dino-park): fixing smooth scrolling for profile in ff nightly

andrew-sunada:
Moving the scroll-behavior rule to the html tag as suggested by the firefox team to make scroll-behavior:smooth work in firefox nightly

Jira: https://jira.mozilla.com/browse/DP-1424

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -82,10 +82,13 @@ export default {
   box-sizing: border-box;
 }
 
+html {
+  scroll-behavior: smooth;
+}
+
 body {
   background: var(--gray-20);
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -82,10 +82,13 @@ export default {
   box-sizing: border-box;
 }
 
+html {
```
fiji-flo: Maybe we wanna add this: https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior#Mind_users_with_motion_sickness

PR #410, opened 2019-07-24
fix(dino-park): fix loading process for org-chart, highlight

andrew-sunada:
Fix the loading times for the org chart and highlight so that the profile highlight shows up either before or starting at the same time as the org chart loading

Jira: https://jira.mozilla.com/browse/DP-1449


src/pages/PageOrgchart.vue
```diff
@@ -198,8 +200,8 @@ export default {
   async created() {
     window.addEventListener('resize', this.updateView);
     this.updateView();
-    await this.fetchData();
-    await this.highlight();
+    this.highlight();
```
fiji-flo: if we run `highlight` before `fetchData` and don't wait on it highlight won't work if the orgchart wasn't loaded before, and if it works it's just a lucky won race condition.

src/pages/PageOrgchart.vue
```diff
@@ -198,8 +200,8 @@ export default {
   async created() {
```
fiji-flo: Just in general, this wouldn't need to be `async` anymore with your changes.

PR #412, opened 2019-07-25
feat(dino-park): adding local time offset difference

andrew-sunada:
Adding local time offset text to profile info.

Jira: https://jira.mozilla.com/browse/DP-1407


src/components/profile/ProfileTeamLocation.vue
```diff
@@ -71,13 +87,58 @@ export default {
...
+        position.coords.longitude,
+      );
+      if (timezone.length === 0) {
+        return;
+      }
+      this.currentTimezone = timezone[0];
+    },
+    processLocationError(error) {
+      console.log('Unable to retrieve location error: ', error.message);
+      this.currentTimezone = this.$store.state.user.timezone.value;
```
fiji-flo: I think we should fallback to the system local time. That's most likely accurate.

src/components/profile/ProfileTeamLocation.vue
```diff
@@ -71,13 +87,58 @@ export default {
...
+        position.coords.latitude,
+        position.coords.longitude,
+      );
+      if (timezone.length === 0) {
+        return;
+      }
+      this.currentTimezone = timezone[0];
+    },
+    processLocationError(error) {
+      console.log('Unable to retrieve location error: ', error.message);
```
fiji-flo: I wouldn't consider denying the location being and error.

src/components/profile/ProfileTeamLocation.vue
```diff
@@ -59,10 +60,25 @@ export default {
...
+      const currentLocalOffset = dt.offset;
+      const printedOffsetRaw = this.getOffsetDiffFromMinutes(
+        currentLocalOffset,
+        profileOffset,
+      );
+      let printedOffset = '';
+      if (printedOffsetRaw !== null && printedOffsetRaw !== 0) {
+        printedOffset = ` | ${
+          printedOffsetRaw > 0 ? '+' : ''
+        }${printedOffsetRaw}hrs to you`;
```
fiji-flo: This renders offset to Nepal like _12.75hrs to you_, could we format it like _12:45hrs_ or _12hrs 45mins_ @mbransn any opinion?

src/components/profile/ProfileTeamLocation.vue
andrew-sunada: @mbransn what sort of behaviour should we expect when the user denies location?

andrew-sunada:
New IAM Jira Story: https://jira.mozilla.com/browse/IAM-173

src/queries/profile.js
```diff
@@ -84,6 +84,12 @@ const USER_MENU_PROFILE = gql`
       primaryEmail {
         value
       }
+      location {
```
fiji-flo: That's not used anymore.

src/components/profile/ProfileTeamLocation.vue
```diff
@@ -59,10 +57,36 @@ export default {
...
+      }).format(currDate);
+      const currentLocalDate = new Intl.DateTimeFormat('en-US', {
+        ...dateOptions,
+        timeZone: this.currentTimezone,
+      }).format(currDate);
+      const hoursDiff =
+        (new Date(profileDate) - new Date(currentLocalDate)) / 36e5;
+      let printedOffset = '';
+      if (hoursDiff !== null && hoursDiff !== 0) {
+        printedOffset = ` | ${hoursDiff > 0 ? '+' : ''}${hoursDiff}hrs to you`;
```
fiji-flo: should we go with +/- here?

src/components/profile/ProfileTeamLocation.vue
```diff
@@ -59,10 +57,36 @@ export default {
...
+      const profileDate = new Intl.DateTimeFormat('en-US', {
+        ...dateOptions,
+        timeZone: this.timezone,
+      }).format(currDate);
+      const currentLocalDate = new Intl.DateTimeFormat('en-US', {
+        ...dateOptions,
+        timeZone: this.currentTimezone,
+      }).format(currDate);
+      const hoursDiff =
+        (new Date(profileDate) - new Date(currentLocalDate)) / 36e5;
```
fiji-flo: nice

src/components/profile/ProfileTeamLocation.vue
```diff
@@ -59,10 +57,36 @@ export default {
...
+      let printedOffset = '';
+      if (hoursDiff !== null && hoursDiff !== 0) {
+        printedOffset = ` | ${hoursDiff > 0 ? '+' : ''}${hoursDiff}hrs to you`;
+      }
       if (this.timezone) {
         return `${this.localtime} local time (${this.getTimezoneName(
```
fiji-flo: Optional: we could change `this.localtime` to just hold the `Date` object with the current time (still updating every second to trigger the computed). And then use `this.localtime` for the offset and to the conversion to the local timezone for display in the `cumputed`.

src/components/profile/ProfileTeamLocation.vue
andrew-sunada: We do. It just doesn't show the '-' here because if the hoursDiff is negative, the - comes with it.

src/components/profile/ProfileTeamLocation.vue
```diff
@@ -58,18 +95,102 @@ export default {
...
+        }
+      }
+
+      // Build local timezone string;
+      const hoursDiff = getHoursDiff(profileDate, currentLocalDate);
+      let printedLocalOffset = '';
+      if (hoursDiff !== null && hoursDiff !== 0) {
+        printedLocalOffset = `${
+          hoursDiff > 0 ? '+' : ''
+        }${hoursDiff}hrs to your local time`;
```
fiji-flo: Still would prefer hh:mm formatting here.

src/assets/js/timezone-utils.js
```diff
@@ -0,0 +1,39 @@
+export const getHoursDiff = (date1, date2) =>
```
fiji-flo: any reason why not to use a function here?

andrew-sunada: What do you mean by this? This is a function...

src/components/profile/ProfileTeamLocation.vue
andrew-sunada: oh good call.

src/components/profile/ProfileTeamLocation.vue
```diff
@@ -58,18 +96,102 @@ export default {
...
+          hoursDiff,
+        )}hrs to your local time`;
+      }
+      if (
+        hoursDiff !== null &&
+        browserHoursDiff !== null &&
+        hoursDiff !== browserHoursDiff
+      ) {
+        return `${printedLocalOffset}${
+          printedBrowserOffset === '' ? '' : `, ${printedBrowserOffset}`
```
fiji-flo: This may result in a leading `, ` if `hoursDiff` is `null` or `0`. Which happens quite frequently if I'm looking at my profile and my system is set to a different time zone (e.g. when traveling).

andrew-sunada: I've accommodated for this in the conditional above.

src/assets/js/timezone-utils.js
```diff
@@ -0,0 +1,56 @@
...
+  if (tz) {
+    options.timeZone = tz;
+  }
+  return new Intl.DateTimeFormat('en-US', options).format(datetime);
+};
+
+export const getBrowserTimezone = () => {
+  try {
+    return Intl.DateTimeFormat().resolvedOptions().timeZone;
+  } catch (e) {
```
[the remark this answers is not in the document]
andrew-sunada: Cool! I've updated the catch block to include a ```console.error```

PR #414, opened 2019-07-29
use bigger picture for user menu

fiji-flo:
This is a intermediate fix to enhance picture quality in the user
menu and top bar. We want to support hidpi picture, which is some
more work.

### DIFF
--- a/src/components/ui/TopBar.vue
+++ b/src/components/ui/TopBar.vue
@@ -77,6 +77,7 @@
                 username: user.primaryUsername.value,
               }"
               :size="40"
+              :pictureSize="100"
               dinoType="Staff"
             ></UserPicture>
           </template>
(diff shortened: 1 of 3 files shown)


src/components/ui/TopBar.vue
```diff
@@ -77,6 +77,7 @@
                 username: user.primaryUsername.value,
               }"
               :size="40"
+              :pictureSize="100"
```
andrew-sunada: In general, I really don't like that we are using hard units here, calculated in the markup. I prefer it to be done in css and to be a percentage unit to be responsive.
That said, that's a bigger battle in this whole codebase.

fiji-flo: Yes, I agree. That to be discussed with UX. However this value is only used to get the actual pictures size (Which should be a `srcset` at some point).

PR #437, opened 2019-08-07
show login in nlx after logout

gdestuynder:
this.. logs you out of sso, see commit msg

also this hardcodes people.mozilla.org - no the biggest deal but is there a way to parametrize this so it's env dependent @fiji?


fiji-flo:
adopted in #442 thanks!

PR #447, opened 2019-08-16
Simplify popover position.

fiji-flo:


andrew-sunada:
This seems fine. Only thing that I noticed was that in production, I see in my account under "contact me" popup, there's both email AND slack, while with this code, i only get Email.
This change wouldn't be the reason for that, but some interesting thing of note..

PR #451, opened 2019-08-21
use srcset for user pictures

LeoMcA:
Testing in Firefox using `layout.css.devPixelsPerPx` suggests Firefox prefers to downscale a higher resolution image than upscale a lower resolution one (which is what we want).

Running `layout.css.devPixelsPerPx` at 2.6x for instance, loads the 264px avatar in the 40px slot even through 100/40px is closer (at 2.5x).

528px not yet included as Firefox doesn't discard 404ing images and use a lower resolution instead, so we need to finish/deploy the changes on the backend first.

### DIFF
--- a/src/components/ui/UserPicture.vue
+++ b/src/components/ui/UserPicture.vue
@@ -80,13 +80,12 @@ export default {
       }
     },
     buildSrcset() {
+      if (!this.slot) return '';
       return [40, 100, 264]
-        .map(
-          (s) =>
-            `${this.avatar.picture}?size=${s} ${(s / this.slot).toPrecision(
-              2,
-            )}x`,
-        )
+        .map((s) => {
+          const res = (s / this.slot).toPrecision(2);
+          return `${this.avatar.picture}?size=${s} ${res}x`;
+        })
         .join(' ');
     },
   },
--- a/src/components/ui/UserPicture.vue
+++ b/src/components/ui/UserPicture.vue
@@ -5,6 +5,7 @@
       ref="img"
       :class="cls"
       :src="src"
+      :srcset="srcset"
       alt=""
       :width="size"
       role="presentation"
@@ -29,7 +30,6 @@ export default {
       },
     },
     cls: String,
-    pictureSize: Number,
     isStaff: Boolean,
   },
   components: {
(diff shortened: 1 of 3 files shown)


src/components/ui/UserPicture.vue
```diff
@@ -77,15 +76,27 @@ export default {
...
       } else {
         this.src = `${this.avatar.picture}?size=${this.slot}`;
+        this.srcset = this.buildSrcset();
       }
     },
+    buildSrcset() {
+      return [40, 100, 264]
+        .map(
+          (s) =>
+            `${this.avatar.picture}?size=${s} ${(s / this.slot).toPrecision(
```
andrew-sunada: Since you can pass any number to ```this.slot```, we should probably do some checks before we divide by 0.

andrew-sunada (changes requested):

src/components/ui/UserPicture.vue
LeoMcA: Good point! I'll get that changed

src/components/ui/UserPicture.vue
```diff
@@ -80,13 +80,12 @@ export default {
...
+      if (!this.slot) return '';
       return [40, 100, 264]
-        .map(
-          (s) =>
-            `${this.avatar.picture}?size=${s} ${(s / this.slot).toPrecision(
-              2,
-            )}x`,
-        )
+        .map((s) => {
+          const res = (s / this.slot).toPrecision(2);
```
andrew-sunada: So, this will still return "NaN" if ```this.slot``` is invalid. What should we do when we get NaN?

andrew-sunada (changes requested):

src/components/ui/UserPicture.vue
LeoMcA: Actually looking closer at the code, I don't think we need to check to see if this.slot is 0/invalid/etc. since it's not a prop and we set it in the `updateSize` method, so it will only ever be 40 or 100 or 264.

andrew-sunada (changes requested):

src/components/ui/UserPicture.vue
fiji-flo: I'd simplify this overall.
Just have something like:
```js
const SRC_SETS = {
  "40": [[40, "1x"], [100, "2x"]],
  "100": [[100, "1x"], [268, "2x"]],
  "264": [[264, "1x"], [528, "2x"]],
}
```
And don't do the math. +1 on using constants for the sizes.

PR #453, opened 2019-08-27
fix(IAM-154): fixing bugs reported for colleagues

andrew-sunada:
Fixing bugs reported for the colleagues story
Jira Story: https://jira.mozilla.com/browse/IAM-154

- Fixing missing colleague view style toggle
-- https://jira.mozilla.com/browse/IAM-235
- Fixing page scroll on clicking of colleague in profile page
-- https://jira.mozilla.com/browse/IAM-236

### DIFF
--- a/src/components/ui/Person.vue
+++ b/src/components/ui/Person.vue
@@ -1,7 +1,10 @@
 <template>
   <div class="person">
     <UserPicture :avatar="{ picture, username }" :size="40" :isStaff="true" />
-    <a v-bind:href="personLink" :title="`${firstName} ${lastName}`">
+    <RouterLink
+      :to="{ name: 'Profile', params: { username } }"
+      :title="`${firstName} ${lastName}`"
+    >
       <div class="person__name-title">
         <div class="person__name">{{ firstName }} {{ lastName }}</div>
         <div class="person__preferred-title">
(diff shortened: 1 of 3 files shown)


src/components/ui/Person.vue
```diff
@@ -1,18 +1,15 @@
 <template>
   <div class="person">
     <UserPicture :avatar="{ picture, username }" :size="40" :isStaff="true" />
-    <RouterLink
-      :to="{ name: 'Profile', params: { username } }"
-      :title="`${firstName} ${lastName}`"
-    >
+    <a v-bind:href="personLink" :title="`${firstName} ${lastName}`">
```
fiji-flo: We should not use `<a>`s.
The page is jumping because the `ShowMore` (my least favorite component) is opened (stays open) when the profile changes. The way this should be handled is setting:
```js
        :closeOnUpdate="true"
        :updateIndicator="JSON.stringify(collection)"
```
in the `ShowMore` component in `ReportingStructureGroup`.
The update indicator is super ugly, maybe we should pass the username down there (or retrieve it from the router, which would be even hackier).
Anyhow, before we clean all of this up I'd be happier with keep the `RouterLink`. Thoughts?

PR #458, opened 2019-08-30
fix(IAM-245): separating office link from officeLocation link

andrew-sunada:
Separating the location link from the office location link

Jira Story: https://jira.mozilla.com/browse/IAM-245


src/components/profile/ProfileTeamLocation.vue
```diff
@@ -258,7 +263,7 @@ export default {
 }
 .profile__team-location a,
 .profile__team-location strong {
-  display: block;
+  display: inline;
```
fiji-flo: This must be `inline-block` otherwise there's no space between location and office:  `LOCATION(OFFICE)` vs `LOCATION (OFFICE)`

PR #459, opened 2019-09-02
Add editing for access groups.

fiji-flo:


src/components/profile/edit/EditMutationWrapper.vue
```diff
@@ -131,25 +130,6 @@ export default {
     return {
       displayProfile: DISPLAY_PROFILE,
       mutateProfile: MUTATE_PROFILE,
-      ...Object.entries(this.initialValues).reduce(
```
fiji-flo: @Gregoor do you remember why that's there?

Gregoor: initialization of those values was fairly long and repetetive, so I opted for this.

src/components/profile/edit/EditAccessGroups.vue
```diff
@@ -0,0 +1,105 @@
...
+        {{ group }}
+      </IconBlock>
+    </IconBlockList>
+    <p v-else>
+      You have not joined any mozilliansorg access group.
+    </p>
+    <a
+      id="access-information__external-link"
+      class="button button--secondary button--action"
+      href="https://mozillians.org"
```
LeoMcA: Have you considered linking to `https://mozillians.org/user/edit/#mygroups` to take a user directly to their groups settings in mozillians.org?

fiji-flo: I considered `https://mozillians.org/user/edit/` but I didn't realize that the groups there are actually links :sweat_smile: 
Thanks.

src/components/profile/edit/EditAccessGroups.vue
```diff
@@ -0,0 +1,105 @@
...
+    PrivacySetting,
+  },
+  methods: {
+    displayLevelsFor,
+  },
+  mounted() {
+    this.$refs.header.focus();
+  },
+  data() {
+    console.log(JSON.stringify(this.initialAccessInformation));
```
andrew-sunada: Did you mean to keep this console log here?

src/components/profile/edit/EditAccessGroups.vue
```diff
@@ -0,0 +1,105 @@
...
+        metadata: { display = DISPLAY_LEVELS.private.value },
+      },
+    } = this.initialAccessInformation;
+    return {
+      editMozilliansorgGroups: {
+        display,
+      },
+      mozilliansorgGroups: [
+        ...Object.entries(
+          this.initialAccessInformation.mozilliansorg.values || {},
```
andrew-sunada: I'm a bit concerned with this code because it assumes some deep values exist that haven't been confirmed or tested for elsewhere in this component. 
```this.initialAccessInformation``` will by default be an empty object. However, here, we are attempting to access deeper values without ever having stated that they may or may not have existed.
The default value of {} "covers" this, so it's not breaking, but if we were to pass an intentional falsey value to this, we would still get an empty object. It just feels a bit hand wavy. Not breaking, but worth mentioning I think.

src/components/profile/edit/EditAccessGroups.vue
fiji-flo: :+1:

src/components/profile/edit/EditAccessGroups.vue
fiji-flo: Actually the default value doesn't cover this. Added safe defaults, I hope you don't mind the destruction assignment with defaults.

PR #461, opened 2019-09-03
Use internal picture api.

fiji-flo:


src/assets/js/avatars.js
```diff
@@ -0,0 +1,5 @@
+export default function avatarUrl(profilePictureUrl) {
+  const url = new URL(profilePictureUrl, window.location.href);
```
andrew-sunada: Seems like we may want to use a more specifically constructed hostname? Maybe not, but just using the window.location.href for the host name feels like a shortcut.

src/components/ui/UserPicture.vue
```diff
@@ -76,7 +77,7 @@ export default {
       ) {
         this.src = await generateIdenticon(this.avatar.username, this.size);
       } else {
-        this.src = `${this.avatar.picture}?size=${this.slot}`;
+        this.src = `${avatarUrl(this.avatar.picture)}?size=${this.slot}`;
```
andrew-sunada: Seems like if we are going to the trouble of building a ```avatarUrl()``` function, we might want to include the query string parameters since that's also part of the url.

src/pages/PageOrgchart.vue
```diff
@@ -145,7 +146,7 @@ function renderNode(node, level = 1) {
     });
   } else {
     img.style.backgroundSize = 'cover';
-    img.style.backgroundImage = `url("${picture}?size=40")`;
+    img.style.backgroundImage = `url("${avatarUrl(picture)}?size=40")`;
```
andrew-sunada: Same as above.

src/assets/js/avatars.js
fiji-flo: `URL` need a full url as second argument passing just a host name errors:
```js
> new URL("/foo", "people.mozilla.com");
> TypeError: people.mozilla.com is not a valid URL.
```

src/components/ui/UserPicture.vue
fiji-flo: :+1: that

PR #462, opened 2019-09-03
fix missing github user name when editing.

fiji-flo:


src/components/profile/edit/EditIdentities.vue
```diff
@@ -164,9 +164,11 @@ export default {
       remove: false,
       display: this.identities.githubDisplay(),
     };
-    this.update();
+    if (this.identities.githubUsername === null) {
```
andrew-sunada: This fix is concerning to me as we are doing logic on fields that we haven't even confirmed exist yet in these components. This is a decent defense for Typescript.

fiji-flo: I do see your concern. However, some trust in JavaScript is okay. Guarding is great, but we're expecting a Identities object here. I made this more explicit by specifying it in the props. But that only results in erroring out earlier. There's only one path to this component in our system which makes it quite traceable what ends up in here.

I do see your point, but some kind of expectations are okay in my opinion.

andrew-sunada: 👍

PR #464, opened 2019-09-09
previewAs

fiji-flo:


src/assets/images/chevron-down-blue-60.svg
```diff
@@ -0,0 +1,13 @@
...
+  height="24"
+  viewBox="0 0 24 24"
+  fill="none"
+  stroke="#0060df"
+  stroke-width="2"
+  stroke-linecap="round"
+  stroke-linejoin="round"
+>
+  <polyline points="6 9 12 15 18 9" />
+</svg>
```
fiji-flo: I don't like adding a new SVG here. But that should be cleaned up in our SVG redo.

src/components/profile/PreviewAs.vue
```diff
@@ -0,0 +1,114 @@
...
+      this.viewAsFilter.active = false;
+      this.viewAsFilter.filter = null;
+    },
+  },
+  destroyed() {
+    this.deactivate();
+  },
+  data() {
+    return {
+      viewAsOptions: [
```
fiji-flo: pretty much the privacy settings. TODO would be rename/structure the icons. Also something for the SVG redo.

fiji-flo:
Ghost border has been removed.

fiji-flo:
@andrew-sunada good catch. I caught the issue: https://github.com/mozilla-iam/dino-park-front-end/pull/464/commits/edf8bf28fadfc4cc605260b4f78503da8c484272

src/pages/PageProfile.vue
```diff
@@ -2,20 +2,19 @@
   <ApolloQuery
     :query="displayProfile"
     :variables="variables"
+    :fetchPolicy="viewAs.filter ? 'no-cache' : 'cache-first'"
     clientId="mutationClient"
   >
     <template slot-scope="{ result: { loading, data, error } }">
+      <PreviewAs v-if="showPreviewAs" :viewAsFilter="viewAs"></PreviewAs>
```
andrew-sunada: I think it might be helpful to put this preview as tag inside the data template. When I pulled loaded this page up, the preview as button came on the page faster than the profile loaded. Seems like you shouldn't be able to interact with previewAs if the profile isn't loaded yet.

https://drive.google.com/open?id=1Piwjn7y97snbMop-QoZOqFRCLOibVjOz

src/pages/PageProfile.vue
```diff
@@ -25,36 +24,62 @@
...
   },
+  mixins: [EnvironmentMixin],
   computed: {
     variables() {
       if (
         this.$route.params.username ===
         this.$store.state.user.primaryUsername.value
       ) {
-        return { username: null };
+        return {
```
andrew-sunada: It seems really messy to me to add an extra data binding to each element that we pass down to each element from this one place. It also seems like the source of this "viewas" functionality doesn't really belong in the PageProfile component. That's more of an opinion that I'm willing to change my mind on, but I think this might belong in it's own mixin.

src/pages/PageProfile.vue
fiji-flo: :+1:

PR #469, opened 2019-09-23
hide things according to scope

fiji-flo:


src/components/profile/ProfileTeamLocation.vue
```diff
@@ -286,6 +284,10 @@ export default {
     display: flex;
     margin: 2em 0;
   }
+  .profile__team-location > * {
```
andrew-sunada: Can we come up with a better way to style the immediate contents of ```profile__team-location```? The ```*``` operator is not a great selector because it has no specificity and it's very fragile for future code. It's also not very descriptive.

src/components/profile/edit/EditPersonalInfo.vue
```diff
@@ -150,33 +154,35 @@
...
-        :value="staffInformation.title.value"
-      />
-      <PrivacySetting
-        label="Title privacy levels"
-        id="field-title-privacy"
-        profileFieldName="staffInformation.title"
-        :profileFieldObject="staffInformation.title"
-        :disabled="true"
-      />
+      <template v-if="scope.isStaff">
```
andrew-sunada: Can we take this moment to refactor this component? We are duplicating a lot of markup here. The fact that we keep having to add ```v-if="scope.isStaff"``` to html that looks almost identical seems like a large indicator we should look at refactoring some of this code.

src/assets/js/scope.js
```diff
@@ -0,0 +1,18 @@
+class Scope {
```
andrew-sunada: Can we add some documentation here to what this file is for? The name is scope, but that term is so overloaded that I don't directly see that it's actually a permissions based handler. I think the ```scope``` in this place is a pretty generic term that doesn't actually tell me what it is, but it's already all over the code, so maybe at least some documentation of what this is meant for so future developers don't try to put stuff in here that doesn't belong here would be great.

PR #471, opened 2019-09-23
feat(IAM-170): adding slack integration to people profile

andrew-sunada:
Adding slack integration to people profile by allowing for slack "signin", and link/username in "contact me" drop down.

Jira Story: https://jira.mozilla.com/browse/IAM-170


src/components/profile/edit/EditSlack.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+  <div :class="{ 'slack-container': true, 'sign-in': slack.v === '' }">
+    <a class="sign-in-link" v-if="slack.v === ''" href="/whoami/slack/add">
+      <img
+        alt="Sign in with Slack"
+        src="https://platform.slack-edge.com/img/sign_in_with_slack.png"
```
fiji-flo: those will be blocked by our CSP headers. Can we self host them?

andrew-sunada: Completed.

andrew-sunada:
> Columns aren't aligned:
> 
> Screenshot from 2019-10-10 19-07-15
> 
> This looks odd, and there's no obvious way to edit my manually entered slack username/turn it into an API-verified one (which I assume we'll display in the same way as other verified accounts?)
> 
> Screenshot from 2019-10-10 19-11-44

Fixed this. You shouldn't be able to edit your username. But the name should have the "Verified" label.

LeoMcA:
But doesn't this mean we're labelling non-verified slack usernames as verified?

Screenshot from 2019-10-11 11-02-10

We need to store somehow whether a slack username is verified or user-set, and if verified show the green text (and style the editing side like the github/bugzilla fields), and if not verified prompt the user when editing to verify their slack account.

We'll probably have to continue to abuse the `#` system in uri names to store this. Perhaps use `VA#` instead of `EA#` when an account is verified? I'm not sure how many knock-on changes this would require.

andrew-sunada:
No, that's a good point. I actually thought that we wouldn't show the slack account unless people had signed in. Basically no grandfathering in. However that was neither discussed, nor makes too much sense. Confirming verification actually won't be hard as the verified values have a very specific format. Good call. I'll make sure that's applied.

LeoMcA:
> We'll probably have to continue to abuse the # system in uri names to store this. Perhaps use VA# instead of EA# when an account is verified? I'm not sure how many knock-on changes this would require.

Thinking about this more, this, and the regex on the uri value, doesn't work and can be easily defeated by either a manual graphql query, or just changing your slack username before the change.

E.g. making my slack username `https://slack.com#robin` gives me:

Screenshot from 2019-10-15 13-32-16

I can't think of a quick solution here which doesn't also require changes to the backend to prevent users modifying certain values in uris. Since time is of the essence, I'd suggest removing the verified label.

We still have some problems with migrating users across to the new system:

Screenshot_2019-10-15 Edit - Profile - Mozilla People Directory - dinopark k8s test sso allizom org

There's no obvious way to link my slack account, or any indication that I could do it. I think we need some kind of help text here, something like: "Turn your slack account into a clickable link by removing it, then adding it again."

I presume if you update your slack username it doesn't get updated on PMO? So once an API-fetched slack uri has been set, we should also show some help text like: "Update your slack username by deleting and adding it again."

/cc @mcroud to see if he has any thoughts on the copy/how to display it

And the use of the user-facing display field, as opposed to a disabled input field, looks weird to me and isn't something we do with any other fields.

mcroud:
Piecing together the information here and on Jira. So the issue is that we want to offer folk a clickable Slack name, much like how it currently works for Discourse.

I'm looking at the following:
image

Is the idea that a user clicks on the **Mozilla Slack ~ Leo** button and is then asked to verify an account? 
The style here is what we see when not logged in and doesn't give me any clues that it is actually clickable or interactive. If I am being asked to click something I think I would need a clear button which says "Verify my slack account".
Am I following right?

LeoMcA:
@mcroud Here's how this PR is currently working for a user with an already-set slack account, and how they get to the "sign in with slack" button.

Peek 2019-10-15 15-10

I don't know enough about the codebase to comment how easy it would be to change this flow, (delete, save, add), so we at least need to prompt the user into this.

LeoMcA:
@andrew-sunada after the standup today we decided we can release this incrementally, and improve the adding UX in a later release. I'll try to review this before the end of the day

andrew-sunada:
> @mcroud Here's how this PR is currently working for a user with an already-set slack account, and how they get to the "sign in with slack" button.
> 
> Peek 2019-10-15 15-10
> 
> I don't know enough about the codebase to comment how easy it would be to change this flow, (delete, save, add), so we at least need to prompt the user into this.

oh wow, I've never had that experience before. how did you acheive that experience?

andrew-sunada:
The intention here was always that we would be in a place where someone would be on the "account editing component" and would receive the "slack sign in" button. If you don't, you'll just have your name show up and it won't be clickable. You can just delete it then and re add it. However I can see that there's no UI showing this. If you've already got a name entered, there should be UI around this. 

So in transition while people still have their original name, we might have a disabled in put that says something like "in order to move forward with this slack user, please delete this slack account, then re add"

PR #472, opened 2019-09-26
fix: fixing scope/nda issue

andrew-sunada:
Fixing bug with scope object reading nda boolean.
Nda value is "null" which isn't "undefined", so things break.

### DIFF
--- a/src/assets/js/scope.js
+++ b/src/assets/js/scope.js
@@ -10,12 +10,20 @@ class Scope {
   }
 
   update(user) {
-    const {
-      staffInformation: { staff: { value: isStaff = false } = {} } = {},
-      accessInformation: {
-        mozilliansorg: { values: { nda: isNdaed = false } = {} } = {},
-      } = {},
-    } = user;
+    let isStaff = false;
+    let isNdaed = false;
+    try {
+      const {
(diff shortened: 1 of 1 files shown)


src/assets/js/scope.js
```diff
@@ -10,12 +10,20 @@ class Scope {
...
+        staffInformation: { staff: { value: isStaffBool = false } = {} } = {},
+        accessInformation: {
+          mozilliansorg: { values: { nda: isNdaedBool = false } = {} } = {},
+        } = {},
+      } = user;
+      isStaff = isStaffBool;
+      isNdaed = isNdaedBool;
+    } catch (e) {
+      console.error('Error: ', e);
+    }
```
fiji-flo: Yeah, that was broken. However, `{ nda: null }` actually means that someone is in the nda group.

```js
    const {
      staffInformation: { staff: { value: isStaff = false } = {} } = {},
      accessInformation: {
        mozilliansorg: { values: { nda: isNdaed = false } = {} } = {},
      } = {},
    } = user || {};
    this.isStaff = isStaff;
    this.isNdaed = isNdaed === null;
``` 

would solve for that and prevent exceptions since every case has a default.

fiji-flo:
resolved in #473

PR #474, opened 2019-10-04
fix search and orgchart for non-staff

fiji-flo:
- hide orgchart button for non-staff in search results
- fall back to username on empy names
- fall back to 'Anonymous User' for default usernames
- show unknown page if orgchart is forbidden


andrew-sunada:
Waiting to get nda permissions on test account to confirm this PR

PR #475, opened 2019-10-04
render identicon in preview-as menu

fiji-flo:


fiji-flo:
Not perfect. @andrew-sunada if you want to change this feel free. But I wanted to unblock this. Ultimately we should move identicons to the back-end. That's why I didn't want to get to creative here.

PR #478, opened 2019-10-09
fix(DP-1465): updating icons

andrew-sunada:
Update the icons of the bugzilla account and the zoom identity

Jira Story: https://jira.mozilla.com/browse/DP-1465


src/components/ui/Icon.vue
```diff
@@ -450,15 +450,9 @@
...
-      <rect
-        fill="currentColor"
-        x="1"
-        y="5"
-        width="15"
-        height="14"
-        rx="2"
-        ry="2"
+      <path
+        id="a"
```
LeoMcA: Not sure this `id` is needed, we should probably also add `fill="currentColor"` like the other svgs in this file.

andrew-sunada: Nice. that's been fixed.

PR #479, opened 2019-10-09
fix(IAM-271): fixing empty message, fixing previewas constants

andrew-sunada:
Fixing empty message so that it shows "No {items} have been added" when not viewing as self, and "You have not added {}" when viewing as self.
Also fixing refactor in PreviewAs.vue

Jira Story: https://jira.mozilla.com/browse/IAM-271

### DIFF
--- a/src/components/profile/PreviewAs.vue
+++ b/src/components/profile/PreviewAs.vue
@@ -32,14 +32,7 @@
 import Icon from '@/components/ui/Icon.vue';
 import Button from '@/components/ui/Button.vue';
 import Select from '@/components/ui/Select.vue';
-
-const PERMISSIONS = {
-  private: 'PRIVATE',
-  staff: 'STAFF',
-  ndaed: 'NDAED',
-  authenticated: 'AUTHENTICATED',
-  public: 'PUBLIC',
-};
+import { DISPLAY_LEVELS } from '@/assets/js/display-levels';
 
 export default {
   name: 'PreviewAs',
@@ -50,7 +43,7 @@ export default {
   },
   methods: {
     activate() {
-      this.viewAs = this.$route.query.pa || PERMISSIONS.private;
+      this.viewAs = this.$route.query.pa || DISPLAY_LEVELS.private.value;
       this.updateFilter();
     },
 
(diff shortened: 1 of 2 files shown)


src/components/profile/PreviewAs.vue
```diff
@@ -79,15 +72,36 @@ export default {
...
+          value: DISPLAY_LEVELS.ndaed.value,
+          icon: DISPLAY_LEVELS.ndaed.icon,
+        },
+        {
+          label: 'Registered',
+          value: DISPLAY_LEVELS.authenticated.value,
+          icon: DISPLAY_LEVELS.authenticated.icon,
+        },
+        {
+          label: 'Public',
```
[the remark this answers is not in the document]
andrew-sunada: I don't disagree with you, but the labels that are in the DISPLAY_LEVELS data structure are not the same as are in this particular map. Specifically the Ndae'd label is different.

src/components/profile/ProfileSection.vue
```diff
@@ -48,6 +49,17 @@ export default {
...
+    emptyMessageText() {
+      if (
+        !this.userOnOwnProfile ||
+        ![undefined, DISPLAY_LEVELS.private.value].includes(
+          this.$route.query.pa,
+        )
+      ) {
+        return this.message;
+      }
+      return this.messageOwn;
```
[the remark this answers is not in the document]
andrew-sunada: Good call here. I'll update this to a little easier stuff to read

src/components/profile/PreviewAs.vue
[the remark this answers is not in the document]
andrew-sunada: If we want to change that text, we should confirm that in the story.

src/components/profile/ProfileSection.vue
```diff
@@ -48,6 +49,16 @@ export default {
...
+    emptyMessageText() {
+      if (
+        !this.userOnOwnProfile ||
+        (this.$route.query.pa !== undefined &&
+          this.$route.query.pa !== DISPLAY_LEVELS.private.value)
+      ) {
+        return this.message;
+      }
+      return this.messageOwn;
+    },
```
LeoMcA: Apologies, this is one of those things that bugs me while I'm brushing my teeth in the morning.

My confusion stemmed from all the negatives, and the need to "cast" an undefined value into private.

I think the positive formulation of this is much more understandable:
```javascript
if (
  this.userOnOwnProfile &&
  DISPLAY_LEVELS.private.value == (this.$route.query.pa || DISPLAY_LEVELS.private.value)
) {
  return this.messageOwn;
}
return this.message;
```

src/assets/js/display-levels.js
```diff
@@ -15,6 +15,11 @@ export const DISPLAY_LEVELS = {
     icon: 'activity',
   },
   ndaed: {
+    label: "NDA'd",
+    value: 'NDAED',
+    icon: 'triangle',
+  },
+  ndaedplusstaff: {
```
LeoMcA: I think this duplication of `ndaed`/`ndaedplusstaff` is the wrong direction to go in. If we *really* want the "view as" drop-down to show something different from the "display level" drop-downs (which I don't think we do, as they're directly linked, and I hope we're reaching that consensus on JIRA), then I think we should just hard-code it here: https://github.com/mozilla-iam/dino-park-front-end/pull/479/commits/d5c531461399fb8c1c846302e393a8c3f1c072ab#diff-056eb55a41bbf1a5f9380763130b1732R91

src/components/profile/PreviewAs.vue
```diff
@@ -79,15 +72,36 @@ export default {
...
+        {
+          label: DISPLAY_LEVELS.authenticated.label,
+          value: DISPLAY_LEVELS.authenticated.value,
+          icon: DISPLAY_LEVELS.authenticated.icon,
+        },
+        {
+          label: DISPLAY_LEVELS.public.label,
+          value: DISPLAY_LEVELS.public.value,
+          icon: DISPLAY_LEVELS.public.icon,
+        },
```
LeoMcA: For staff, authenticated, public (and hopefully nda), we can just include the whole `DISPLAY_LEVELS.foobar` object.

src/assets/js/display-levels.js
andrew-sunada: Yea, that makes good sense. Thanks for that catch!

PR #498, opened 2019-11-01
Don't show keys edit interface to non-LDAP accounts

LeoMcA:
https://jira.mozilla.com/browse/IAM-333
https://github.com/mozilla-iam/dino-park-issues/issues/168

### DIFF
--- a/src/assets/js/scope.js
+++ b/src/assets/js/scope.js
@@ -14,10 +14,12 @@ class Scope {
       staffInformation: { staff: { value: isStaff = false } = {} } = {},
       accessInformation: {
         mozilliansorg: { values: mozilliansorgGroups = {} } = {},
+        ldap: { values: ldapGroups = {} } = {},
       } = {},
     } = user || {};
     this.isStaff = isStaff;
     this.isNdaed = 'nda' in (mozilliansorgGroups || {});
+    this.isLdap = ldapGroups ? Object.keys(ldapGroups).length > 0 : false;
   }
 }
 
--- a/src/assets/js/scope.js
+++ b/src/assets/js/scope.js
@@ -19,7 +19,7 @@ class Scope {
     } = user || {};
     this.isStaff = isStaff;
     this.isNdaed = 'nda' in (mozilliansorgGroups || {});
-    this.isLdap = ldapGroups ? Object.keys(ldapGroups).length > 0 : false;
+    this.isLdap = Boolean(ldapGroups);
   }
 }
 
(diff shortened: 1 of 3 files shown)


src/assets/js/scope.js
```diff
@@ -14,10 +14,12 @@ class Scope {
       staffInformation: { staff: { value: isStaff = false } = {} } = {},
       accessInformation: {
         mozilliansorg: { values: mozilliansorgGroups = {} } = {},
+        ldap: { values: ldapGroups = {} } = {},
       } = {},
     } = user || {};
     this.isStaff = isStaff;
     this.isNdaed = 'nda' in (mozilliansorgGroups || {});
+    this.isLdap = ldapGroups ? Object.keys(ldapGroups).length > 0 : false;
```
fiji-flo: `ldapGroups` will be `null` for non LDAP users. `Object.keys(ldapGroups)` will error there. That's non-intuitive behavior of the destructing assignment, if something is `null` the default does not apply.

But we can leverage this and just check if `ldapGroups === null` in the ternary. LDAP is the only allowed publisher of that field and if it's not null that user has LDAP.

src/components/profile/Profile.vue
```diff
@@ -193,6 +193,8 @@
       :empty="!sections.keys"
       message="No keys have been added yet."
       messageOwn="You haven't added any keys yet."
+      messageNoLdap="You need an LDAP account to add keys."
```
fiji-flo: I'd opt for a different language. _Need_ sounds like you're missing out.

src/assets/js/scope.js
fiji-flo: `this.isLdap = ldapGroups !== null;` or `this.isLdap = Boolean(ldapGroups);`  that's what it could be.

LeoMcA: > ldapGroups will be null for non LDAP users. Object.keys(ldapGroups) will error there.

Since `!!null === false` then `Object.keys(ldapGroups)` will never get executed, no?

But cool, wasn't sure if it could ever be an empty object, hence the ternary, so I'll remove it.

src/components/profile/Profile.vue
LeoMcA: "Without an LDAP account you can't add keys."?

"You can't yet add keys without an LDAP account."?

fiji-flo: _Editing keys without an LDAP account not supported yet._?

PR #501, opened 2019-11-04
limit tags allowed in bio to prevent faking other fields

LeoMcA:


src/components/profile/ProfileDescription.vue
```diff
@@ -9,6 +9,8 @@
 import marked from 'marked';
 import insane from 'insane';
 
+insane.defaults.allowedTags = ['a', 'em', 'li', 'ol', 'p', 'strong', 'ul'];
```
fiji-flo: Do you know how this translates into markdown?

LeoMcA: As in, what markdown we're supporting?

It's just:

`[a](http://example.com)`
`*em*`
```
* <ul>
* <li>
```
```
1. <ol>
2. <li>
```
```
p

p
```
`**strong**`

fiji-flo: maybe add `pre` and `code` for \`\`\` and \`

LeoMcA:
@fiji-flo since we now have the border surrounding the bio, I decided to be much more permissive here (while still restricting things like images).

The full list of supported tags can be seen here:

https://dinopark.k8s.test.sso.allizom.org/p/leo

(at least until my profile gets overwritten, so here's the source for posterity:)

    [a](http://example.com)

    > blockquote

    `code`

    ~del~

    *em*

    # h1

    ## h2

    ### h3

    #### h4

    ##### h5

    ###### h6

    1. ol
    1. li

    ```
    pre?
    ```

    <small>small</small>

    **strong**

    <sub>sub</sub>

    <sup>sup</sup>

    * ul
    * li

PR #503, opened 2019-11-04
hide dino type in edit screen

LeoMcA:
https://jira.mozilla.com/browse/IAM-335?focusedCommentId=31976&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-31976

### DIFF
--- a/src/components/profile/edit/EditPersonalInfo.vue
+++ b/src/components/profile/edit/EditPersonalInfo.vue
@@ -30,11 +30,7 @@
     </header>
     <div class="edit-personal-info">
       <div class="edit-personal-info__picture">
-        <UserPicture
-          :avatar="avatar"
-          :size="264"
-          :isStaff="scope.isStaff"
-        ></UserPicture>
+        <UserPicture :avatar="avatar" :size="264"></UserPicture>
         <button
           class="button button--icon-only edit-personal-info__picture-edit-button"
           type="button"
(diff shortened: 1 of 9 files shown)


src/components/profile/edit/EditPersonalInfo.vue
```diff
@@ -34,6 +34,7 @@
           :avatar="avatar"
           :size="264"
           :isStaff="scope.isStaff"
+          :edit="true"
```
andrew-sunada: Why aren't we just removing this icon entirely from the editpersonalinfo.vue file? If we are hiding the picture always, why do we need to add a boolean flag to the userpicture?
Also, it doesn't seem appropriate that the UserPicture knows anything about any editing mechanism of the profile at the moment.

fiji-flo: We could also rename the `isStaff` attribute into `showLabel` and just use that for everything. Thoughts?

LeoMcA:
Do we want the "m" to appear on your usericon in the top right as well? There was a legacy `dinoType="Staff"` line in `TopBar.vue`.

fiji-flo:
> Do we want the "m" to appear on your usericon in the top right as well? There was a legacy `dinoType="Staff"` line in `TopBar.vue`.

Yes I think it's nice to display it there. It helps me to figure out if I'm logged in as staff or volunteer.

PR #507, opened 2019-11-08
hide staff only cards in preview

fiji-flo:


src/pages/PageProfile.vue
```diff
@@ -73,6 +74,13 @@ export default {
     showPreviewAs() {
       return this.variables.username === null && this.editing === null;
     },
+    allowStaffOnlyFields() {
+      console.log(this.viewAs.filter);
```
LeoMcA: Think you missed this, otherwise, looks good :+1:

fiji-flo: :+1:

PR #508, opened 2019-11-08
add the data classifications a user has access to to their profile

LeoMcA:
https://jira.mozilla.com/browse/IAM-338

Waiting for feedback in JIRA issue before merging.

### DIFF
--- a/src/components/profile/view/ViewPersonalInfo.vue
+++ b/src/components/profile/view/ViewPersonalInfo.vue
@@ -50,21 +50,30 @@
         :officeLocation="staffInformation.officeLocation.value"
         :timezone="timezone.value"
       ></ProfileTeamLocation>
-      <MetaList v-if="this.staffInformation.staff.value">
+      <MetaList>
         <h3 class="visually-hidden">Meta</h3>
-        <Meta
+        <MetaItem metaKey="Access to">
+          <DataClassification
+            :staffInformation="staffInformation"
+            :accessInformation="accessInformation"
+          />
+        </MetaItem>
+        <MetaItem
           metaKey="Worker type"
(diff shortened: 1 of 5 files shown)


src/components/profile/view/ViewPersonalInfo.vue
```diff
@@ -50,21 +50,30 @@
         :officeLocation="staffInformation.officeLocation.value"
         :timezone="timezone.value"
       ></ProfileTeamLocation>
-      <MetaList v-if="this.staffInformation.staff.value">
+      <MetaList>
```
fiji-flo: Doesn't wrap on mobile (this was never good, but now gets worse). Maybe something like helps:
```css
display: grid;
grid-template-columns: repeat(auto-fill,minmax(10em, 1fr));
grid-gap: 1em;
```

LeoMcA: I pretty much went with this, just reduced the min a little, and used `auto-fit` so that the full width was always used.

jens1o:
As I can't find an public issue for this: Maybe you could include `Slack` if the person is not a staff member, which would make it easier to check whether the person should have access, like mhoye did this week.

LeoMcA:
@jens1o is the group allowed access to slack different from the general nda group?

We've talked internally a little bit about adding further classifications here (and more of a sense of what this actually gives you access to, rather than just what classifications of data you have access to), but don't have the resources to do this in the immediate future. Can you file an issue for this under https://github.com/mozilla-iam/dino-park-issues/issues (even if it's just a copy paste of your comment here) - it's always nice to have actual requests for features rather than us just hypothesising about what users might find useful (and it'll be easier to track). Thanks!

PR #509, opened 2019-11-11
guard null key fields

fiji-flo:


fiji-flo:
Hot fixing test. Happy to revisit later.

PR #510, opened 2019-11-14
reload via page to prevent auth issues

fiji-flo:


src/components/profile/edit/EditPersonalInfo.vue
```diff
@@ -405,7 +405,7 @@ import EditPictureModal from './EditPictureModal.vue';
 import Fetcher from '@/assets/js/fetcher';
 
 // TODO: Why do we need failoverOnError here?
```
LeoMcA: This can probably be removed

src/pages/PageSearchResult.vue
```diff
@@ -84,7 +84,10 @@ import SearchResultList from '@/components/search/SearchResultList.vue';
 import SearchToggle from '@/components/search/SearchToggle.vue';
 import Fetcher from '@/assets/js/fetcher';
 
-const fetcher = new Fetcher({ failoverOn: [302] });
+const fetcher = new Fetcher({
+  isError: (e) =>
+    e instanceof TypeError && e.message.startsWith('NetworkError'),
```
LeoMcA: Will search throw errors which aren't NetworkErrors? Will other fetches throw errors which aren't NetworkErrors? If they do, do we want them to reload on all fetch errors, but only search to reload on NetworkErrors?

public/reload.html
```diff
@@ -0,0 +1,38 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width,initial-scale=1.0" />
+    <link
+      href="https://fonts.googleapis.com/css?family=Open+Sans:400,700|Zilla+Slab"
```
LeoMcA: It's probably cached already, but do we really need open sans on a page which users will hopefully see for as little time as possible.

src/main.js
```diff
@@ -33,7 +34,7 @@ function errorHandler(error) {
       networkError.message.startsWith('NetworkError')) ||
       failoverOn.includes(networkError.statusCode))
```
LeoMcA: Do we want apollo queries to only reload on 302s, when we've removed that restriction everywhere else?

fiji-flo: Not happy to write this: I'm to scared to change this right now.

src/pages/PageSearchResult.vue
fiji-flo: Mostly keeping previous behavior to avoid breaking things for now.

public/reload.html
fiji-flo: :+1:

PR #512, opened 2019-11-15
first step towards using fluent for i18n

LeoMcA:
add library and helper functions, use it to extract and update
strings for edit page info icons

https://jira.mozilla.com/browse/IAM-362


src/assets/js/fluent.js
```diff
@@ -0,0 +1,79 @@
...
+
+  static sanitize(message, tags = {}) {
+    return insane(message, {
+      filter(token) {
+        const name = token.attrs['data-l10n-name'];
+        if (Object.keys(tags).includes(name) && tags[name].tag === token.tag) {
+          Object.keys(token.attrs).forEach((k) => {
+            if (k !== 'title') delete token.attrs[k];
+          });
+          Object.entries(tags[name]).forEach(([k, v]) => {
```
andrew-sunada: Not sure how many items this might have, but as a point of reference, I usually go with for loop since it's the faster iterator: https://hackernoon.com/javascript-performance-test-for-vs-for-each-vs-map-reduce-filter-find-32c1113f19d7

Not that it's that big of a deal necessarily depending on how many tags you'll have but, just a thought.

src/assets/js/fluent.js
```diff
@@ -0,0 +1,79 @@
...
+          return true;
+        }
+        return false;
+      },
+    });
+  }
+
+  getMessage(id, attr = null, args = {}) {
+    const message = this.bundle.getMessage(id);
+
```
andrew-sunada: So one thing that I think that might make this a bit simpler to read. (Again, entirely opinionated) Is to put the fail cases first so I have as little of the functionality in deep logic blocks as possible. I think the following might be a bit more readable:

```const message = this.bundle.getMessage(id);
if(!message){
console.error(`string ${id} doesn't exist`);
return '';
}
      if (attr) {
        const messageAttr = message.attributes[attr];
        if (messageAttr) {
          const errors = [];
          const msg = this.bundle.formatPattern(messageAttr, args, errors);
          if (errors.length) {
            console.error(errors);
          }
          return msg;
        }
        console.error(`string ${id} with ${attr} attribute doesn't exist`);
      }
if (message.value) {
        return this.bundle.formatPattern(message.value);
      }```

Again, small change, but I feel like it makes it a bit more readable.

File: src/components/Fluent.vue [hunk 3]
```diff
@@ -0,0 +1,25 @@
...
+export default {
+  name: 'Fluent',
+  props: {
+    id: String,
+    attr: String,
+    args: Object,
+    tags: Object,
+  },
+  computed: {
+    getTranslation() {
```
andrew-sunada: Can we get this unit tested?

src/components/profile/edit/EditPersonalInfo.vue
```diff
@@ -403,6 +461,7 @@ import Tooltip from '@/components/ui/Tooltip.vue';
 import UserPicture from '@/components/ui/UserPicture.vue';
 import EditPictureModal from './EditPictureModal.vue';
 import Fetcher from '@/assets/js/fetcher';
+import Fluent from '@/components/Fluent.vue';
 
 // TODO: Why do we need failoverOnError here?
 const fetcher = new Fetcher({ failoverOn: [302], failoverOnError: true });
```
andrew-sunada: I don't think we do? We also should move it even if we do to be out of the component.

src/main.js
```diff
@@ -105,6 +107,11 @@ const store = new Vuex.Store({
 });
 
 Vue.mixin({
+  methods: {
+    fluent(...args) {
```
andrew-sunada: Out of curiosity, what's the argument behind putting this in the global methods mixin, rather than in the /components/_mixins/ directory?

src/main.js
```diff
@@ -78,6 +79,7 @@ const store = new Vuex.Store({
   state: {
     user: null,
     scope: new Scope(),
+    fluent: new Fluent(),
```
andrew-sunada: The state is usually reserved for storing application state. I don't know Fluent too well, but I'm wondering if a better place for the Fluent instance might be just another Fluent.js file that we import from in the rest of the application. I can't quite tell if this is implemented via a technical necessity, but I'm just wondering if that's possible.

LeoMcA: This was one of the things I wanted to discuss: so, we have to do some setup with fluent (which you'll see in the `Fluent` constructor), like parsing the string files and constructing the bundle out of it. My understanding is, if we import and initialize a `Fluent` class in every file we want to use it (which will eventually be close to all of them), we'll be doing this setup multiple times, for no real reason. We might also want to cache computed translated strings in the future.

Maybe there's some better way of doing "initialize once, use everywhere" in Vue? I just followed the existing pattern we're using for `Scope`.

src/main.js
LeoMcA: Eventually we'll be using this in every file with user-facing strings (which will be close to all of them)

src/components/profile/edit/EditPersonalInfo.vue
LeoMcA: Is this referencing `TODO: Why do we need failoverOnError here?` Florian removed this in a separate PR: https://github.com/mozilla-iam/dino-park-front-end/pull/510/files#diff-65e12fed037aa326de35fbf8f19bb04fL407

andrew-sunada: Wonderful!

src/main.js
andrew-sunada: Oh yea, that makes a lot of sense. So setting up the Fluent instantiating in one file and then importing it every other file would be the way to handle that. It doesn't reinstantiate the file on each import.
I'm doing that actually for the store and the router in my PR: https://github.com/mozilla-iam/dino-park-front-end/pull/505
You can prove this to yourself by putting a console.log in the "fetcher.js" file which is being pulled in by multiple components and confirm that file gets run only once.

The difference between this and the Scope constructor is that the Scope is actually storing the state of the user if you look at the scope object. So it makes a little more sense than storing a library.
Super cool stuff! Thanks for getting this in our codebase!

src/main.js
andrew-sunada: Ah that makes sense. Cool.

src/assets/js/fluent.js
```diff
@@ -0,0 +1,79 @@
...
+    return Fluent.sanitize(message, tags);
+  }
+
+  static sanitize(message, tags = {}) {
+    return insane(message, {
+      filter(token) {
+        const name = token.attrs['data-l10n-name'];
+        if (Object.keys(tags).includes(name) && tags[name].tag === token.tag) {
+          Object.keys(token.attrs).forEach((k) => {
+            if (k !== 'title') delete token.attrs[k];
```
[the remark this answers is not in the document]
LeoMcA: Agreed, I thought the linter would have caught this, but apparently not - I've removed this section of code anyway because of some bugs I found while writing tests

src/assets/js/fluent.js
LeoMcA: Oh man I've now read way too many articles on for vs forEach vs for...in vs for...of and I think I'm more confused than when I started!

I like for...of (readability wise it's the same as forEach, at least in my eyes), but our linter currently prohibits it: `iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations (no-restricted-syntax)`

src/assets/js/fluent.js
```diff
@@ -0,0 +1,79 @@
...
+        case 2:
+          if (typeof parameters[1] === 'string') {
+            return this.get({ id: parameters[0], attr: parameters[1] });
+          }
+          return this.get({ id: parameters[0], args: parameters[1] });
+        default:
+          return this.get({ id: parameters[0] });
+      }
+    }
+    const { id, attr, args, tags } = parameters[0];
```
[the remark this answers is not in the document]
LeoMcA: I'm not sure if this is a problem - `parameters.length == 0` only if the function has been called with no arguments, which isn't something we want/need to support. Probably better that it blows up than attempts to salvage that case.

src/assets/js/fluent.js
```diff
@@ -0,0 +1,79 @@
...
+
+    this.bundle = bundle;
+  }
+
+  get(...parameters) {
+    if (typeof parameters[0] === 'string') {
+      switch (parameters.length) {
+        case 2:
+          if (typeof parameters[1] === 'string') {
+            return this.get({ id: parameters[0], attr: parameters[1] });
```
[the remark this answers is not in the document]
LeoMcA: I've DRYed up the multiple `id: parameters[0]`, but I generally like to avoid mutations where I can.

src/main.js
LeoMcA: I still want to avoid having to import fluent in every file (because we'll eventually be using it in close to every one), but yes agreed that it makes no sense to be in the vuex store. I've achieved that by just initializing Fluent as a const in `main.js` and still referencing it in the mixin method.

src/assets/js/fluent.js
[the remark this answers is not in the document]
andrew-sunada: Fairplay. Not my favourite way of handling that, but I get your point.

PR #513, opened 2019-11-19
support editing display level for non pmo fields

fiji-flo:


src/components/profile/edit/EditPersonalInfo.vue
```diff
@@ -11,6 +11,8 @@
       location,
       timezone,
       description,
+      staffInformationTitle: staffInformationTitle.display,
+      staffInformationOfficeLocation: staffInformationOfficeLocation.display,
```
LeoMcA: This could be confusing in future, but there's only one instance of it. Just how easy is it to update the mutation variable name to `staffInformationTitleDisplay` and `staffInformationOfficeLocationDisplay` on the client side?

src/assets/js/display-levels.js
```diff
@@ -47,14 +47,24 @@ const DISPLAY_ANY = [
...
+  primaryEmail: DISPLAY_ANY,
+};
+
+export function displayLevelsFor(field, scope = null) {
+  return (
+    (field &&
+      ((scope && !scope.isStaff && NON_STAFF_DISPLAY_LEVELS[field]) ||
+        VALID_DISPLAY_LEVELS[field])) ||
+    DISPLAY_NOT_PRIVATE
+  );
```
LeoMcA: This makes my brain hurt, but making it clearer without making it massively more verbose seems quite difficult. I think we can get rid of some brackets, can't we? And isn't the `field &&` guard unnecessary since `{}[undefined] == undefined`?

I _think_ `scope && !scope.isStaff && NON_STAFF_DISPLAY_LEVELS[field] || VALID_DISPLAY_LEVELS[field] || DISPLAY_NOT_PRIVATE` is logically equivalent.

fiji-flo: Did that, thanks. Formatter added some parenthesis.

src/components/profile/edit/EditPersonalInfo.vue
fiji-flo: I'll actually also update the back-end. You're right here.

PR #514, opened 2019-11-19
fix popover on mobile

fiji-flo:


fiji-flo:
Please take a look. Merging so QA can take a look on mobile.

LeoMcA:
:+1:

PR #520, opened 2019-11-21
Setup access groups

andrew-sunada:
Getting new demo stuff for viewing and editing access-groups on to the demo branch


src/components/access_group/AccessGroupMembers.vue
```diff
@@ -0,0 +1,268 @@
...
+          :class="{ 'tabs-container__item': true, active: isTabActive(tab) }"
+          v-for="(tab, idx) in tabList"
+          :key="idx"
+          @click="handleTabClick(tab)"
+        >
+          {{ tab.label }}
+        </li>
+      </ul>
+    </nav>
+    <ul class="members-container__list">
```
fiji-flo: We can add a sort by role (i.e. curators are returned in the first pages) to the API. Then, and also in general I think this should just be a flex container and not a list of lists. Right now it breaks in mobile (the lists show underneath each other an curators are separated).

.prettierrc
```diff
@@ -1,5 +1,7 @@
 {
+  "printWidth": 90,
```
fiji-flo: any specific reason, besides preference for this: https://prettier.io/docs/en/options.html#print-width

.prettierrc
```diff
@@ -1,5 +1,7 @@
...
   "singleQuote": true,
-  "trailingComma": "all",
-  "arrowParens": "always"
+  "trailingComma": "es5",
+  "jsxBracketSameLine": true,
+  "jsxSingleQuote": true
```
fiji-flo: All of these were added by Gregor. I prefer just using defaults as much as possible.
Do the jsx rules actually apply to Vue? Is see double quotes in the HTML in this PR so I'm asking what impact does this have. Same for same line bracket. I'd again prefer the defaults here. Brackets for less line change and double quotes in HTML since it's what the Firefox inspector will make out of it anyway.

src/components/access_group/AccessGroupMembers.vue
andrew-sunada: I agree that this should be a flex box, but @HerminaC had requested that the whole row not drop when one expands. That was the way I could think of to make this happen. But I can see how this messes things up in mobile. I'll rethink.

.prettierrc
andrew-sunada: Nope. I'll remove.

.prettierrc
andrew-sunada: I'll remove these.

PR #538, opened 2019-12-04
add permanent scrollbar to stop content jumping around

LeoMcA:
when navigating between profiles

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -82,6 +82,10 @@ export default {
   box-sizing: border-box;
 }
 
+html {
+  overflow-y: scroll;
+}
+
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -88,6 +88,7 @@ body {
   margin: 0;
   font-size: 87.5%;
   letter-spacing: 0.04em;
+  overflow-y: scroll;
```
fiji-flo: this must be set for `html` to work afaik.

LeoMcA: it worked for me on body, but `html` seems to be the safer option

PR #543, opened 2019-12-12
add info icons for general contributor audience

LeoMcA:
https://jira.mozilla.com/browse/IAM-392

### DIFF
--- a/src/components/profile/edit/EditIdentities.vue
+++ b/src/components/profile/edit/EditIdentities.vue
@@ -9,7 +9,14 @@
     formName="Edit identities"
   >
     <header class="profile__section-header" ref="header" tabindex="-1">
-      <h2>Identities</h2>
+      <div class="profile__section-header__title-info">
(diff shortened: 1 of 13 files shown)


src/components/profile/edit/EditIdentities.vue
```diff
@@ -9,7 +9,14 @@
     formName="Edit identities"
   >
     <header class="profile__section-header" ref="header" tabindex="-1">
-      <h2>Identities</h2>
+      <div class="profile__section-header__title-info">
```
fiji-flo: In my _we don't need divs_ mode: if we change the `div` to an `h2` and put the _title_ directly in that. We just need to change the display of the `Tooltip` to `inline-flex` and it would look that same I guess.

LeoMcA: I originally tried without the div, but the css got very messy - the header already has flex set which needs to be overcome if the tooltip is outside the <h2>, and if it's inside then we need to unset all the h2 css on the tooltip.

PR #544, opened 2019-12-16
replace column css with grid on icon block lists

LeoMcA:


src/components/ui/IconBlockList.vue
```diff
@@ -15,9 +15,11 @@ export default {
   padding-left: 0em;
 }
 @media (min-width: 50em) {
```
fiji-flo: how about we use `grid-template-colums: 1fr` for mobile?

PR #546, opened 2019-12-16
add links to public discourse posts

LeoMcA:


src/assets/js/fluent.js
```diff
@@ -57,6 +61,9 @@ class Fluent {
             });
             return true;
           }
+          if (inlineTags.includes(token.tag)) {
```
andrew-sunada: <Opinion> I prefer to have all conditional returns of the same value under the same conditional. I feel like this conditional and the one above are super related and can be refactored to include one ```return true;``` statement

src/components/ui/Footer.vue
```diff
@@ -37,7 +37,10 @@ export default {
           abbr: 'Community Participation Guidelines',
         },
         {
-          link: this.globalLinks.faq,
+          link:
+            this.scope.isStaff || this.scope.isNdaed
+              ? this.globalLinks.faq
```
andrew-sunada: This is a bit nitpicky, but if this ```faq``` field is the 'private' field, can we just make it be ```faqPrivate``` for clarity sake?

src/pages/PageHome.vue
```diff
@@ -74,38 +74,43 @@ export default {
       return this.$store.state.user.primaryUsername.value;
     },
     cards() {
-      const getLink = (url, text) => {
-        return `<a href="${url}" target="_blank" rel="noopener noreferrer">${text}</a>`;
-      };
+      const privacyLink =
+        this.scope.isStaff || this.scope.isNdaed
+          ? this.globalLinks.profilePrivacy
```
andrew-sunada: Same as above with the 'private' vs 'public'

src/components/ui/Footer.vue
LeoMcA: Agreed, I left it as just `faq` as that's what it was already called, but that doesn't make much sense to fresh eyes. I think I'll probably call it `faqNda` - as that's the visibility of the post in Discourse.

src/assets/js/fluent.js
LeoMcA: That's true, I hacked this in without thinking about it too much, but now I'm looking at it again it's (kinda) possible that as written this would allow translators to add unintended attributes to these whitelisted `inlineTags` (which we don't want). I'll rewrite this :)

LeoMcA: Actually I discovered this couldn't happen, because of the sanitizing that the library was already doing, but I've added some tests to verify this.

PR #571, opened 2020-03-12
use membership / curator status from API

fiji-flo:


fiji-flo:
Yes. I saw the original intend and that you solved it somewhere else :+1:

PR #572, opened 2020-03-12
Prepare for public 1

fiji-flo:
depends on #571

### DIFF
--- a/src/assets/js/scope.js
+++ b/src/assets/js/scope.js
@@ -11,15 +11,21 @@ class Scope {
 
   update(user) {
     const {
+      uuid: { value: uuid = '00000000-0000-0000-0000-000000000000' } = {},
+      primaryUsername: { value: username = '' } = {},
       staffInformation: { staff: { value: isStaff = false } = {} } = {},
       accessInformation: {
         mozilliansorg: { values: mozilliansorgGroups = {} } = {},
         ldap: { values: ldapGroups = {} } = {},
       } = {},
     } = user || {};
+    this.username = username;
+    this.uuid = uuid;
+    this.isLoggedIn = Boolean(username);
     this.isStaff = isStaff;
     this.isNdaed = 'nda' in (mozilliansorgGroups || {});
     this.isLdap = Boolean(ldapGroups);
+    this.isGroupCreator = 'group_creators' in (mozilliansorgGroups || {});
   }
 }
 
--- a/src/components/profile/edit/EditAccessGroups.vue
+++ b/src/components/profile/edit/EditAccessGroups.vue
@@ -94,7 +94,7 @@
             name: 'Create Access Group',
           }"
           rel="noreferrer noopener"
-          v-if="getFeature('accessGroupsToggle')"
+          v-if="getFeature('accessGroupsToggle') && showEditAccessGroups"
         >
           <Icon id="plus" :width="18" :height="18" />{{
             fluent('profile_access-groups_mozillians', 'create-access-groups')
@@ -106,7 +106,6 @@
           href="https://mozillians.org/user/edit/#mygroups"
           target="_blank"
           rel="noreferrer noopener"
-          v-else
         >
           <Icon id="external" :width="18" :height="18" />
           {{ fluent('profile_access-groups_mozillians', 'edit') }}
@@ -147,6 +146,11 @@ export default {
   },
   mounted() {
     this.$refs.header.focus();
+  },
+  computed: {
+    showEditAccessGroups() {
+      return this.$store.state.scope.isGroupCreator;
+    },
   },
   data() {
     const {
--- a/src/components/profile/edit/EditMutationWrapper.vue
+++ b/src/components/profile/edit/EditMutationWrapper.vue
@@ -12,7 +12,7 @@
         class="mutation-form"
         action=""
         :novalidate="novalidate"
-        v-on:submit.prevent="(ev) => check(mutate, ev)"
+        v-on:submit.prevent="ev => check(mutate, ev)"
         :aria-label="formName"
       >
         <slot></slot>
@@ -93,7 +93,7 @@ export default {
       store,
       {
         data: { profile },
-      },
+      }
     ) {
       const data = store.readQuery({
         query: DISPLAY_PROFILE,
@@ -125,7 +125,7 @@ export default {
   },
   computed: {
     loggedInUser() {
-      return this.$store.state.user.primaryUsername.value;
+      return this.$store.state.scope.username;
     },
   },
   data() {
--- a/src/components/profile/edit/EditPersonalInfo.vue
+++ b/src/components/profile/edit/EditPersonalInfo.vue
@@ -22,7 +22,7 @@
       v-bind="{
         picture,
         pictureData,
-        username: loggedInUser.primaryUsername,
+        username: loggedInUsername,
         staffInformation,
       }"
       @close="showPictureModal = false"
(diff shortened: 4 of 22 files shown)


fiji-flo:
I can redo this without formatting tomorrow :man_shrugging:

src/assets/js/scope.js
```diff
@@ -11,15 +11,21 @@ class Scope {
...
         ldap: { values: ldapGroups = {} } = {},
       } = {},
     } = user || {};
+    this.username = username;
+    this.uuid = uuid;
+    this.isLoggedIn = Boolean(username);
     this.isStaff = isStaff;
     this.isNdaed = 'nda' in (mozilliansorgGroups || {});
     this.isLdap = Boolean(ldapGroups);
+    this.isGroupCreator = 'group_creators' in (mozilliansorgGroups || {});
```
andrew-sunada: This is fine for now, but I'd like to see this abstracted a bit more away. Like maybe have a config file of keys or something. A static string like this (and like the nda one)  seems like it could use some refactoring. Not necessary for now, but worth mentioning.

src/components/profile/edit/EditPersonalInfo.vue
```diff
@@ -503,13 +503,13 @@ export default {
     Fluent,
   },
   computed: {
-    loggedInUser() {
-      return this.$store.state.user;
+    loggedInUsername() {
+      return this.$store.state.scope.username;
```
andrew-sunada: If we are just using this state.scope situation, we should really be using vuex mapGetters

src/components/profile/edit/EditMutationWrapper.vue
```diff
@@ -125,7 +125,7 @@ export default {
   },
   computed: {
     loggedInUser() {
-      return this.$store.state.user.primaryUsername.value;
+      return this.$store.state.scope.username;
```
andrew-sunada: Also using mapGetters

src/components/profile/edit/EditAccessGroups.vue
```diff
@@ -148,6 +147,11 @@ export default {
   mounted() {
     this.$refs.header.focus();
   },
+  computed: {
+    showEditAccessGroups() {
+      return this.$store.state.scope.isGroupCreator;
```
andrew-sunada: We should be using mapGetters here.

src/components/ui/Icon.vue
```diff
@@ -53,25 +53,23 @@ export default {
     id: String,
   },
   mounted() {
-    generateIdenticon(this.$store.state.user.primaryUsername.value, 40).then(
-      a => {
-        this.generatedAvatarUrl = a;
-      }
-    );
+    generateIdenticon(this.$store.state.scope.username, 40).then(a => {
```
andrew-sunada: We should be using mapGetters here.

src/components/ui/Icon.vue
```diff
@@ -53,25 +53,23 @@ export default {
...
+      this.generatedAvatarUrl = a;
+    });
   },
   computed: {
     computedAvatarUrl() {
-      if (this.$store.state.user.picture.value) {
+      if (
+        this.$store.state.scope.isLoggedIn &&
+        this.$store.state.user.picture.value
```
andrew-sunada: mapGetters

src/components/ui/Icon.vue
```diff
@@ -53,25 +53,23 @@ export default {
...
           this.$store.state.user.picture.value,
           40,
           true
         );
       } else {
-        generateIdenticon(
-          this.$store.state.user.primaryUsername.value,
-          40
-        ).then(a => {
+        generateIdenticon(this.$store.state.scope.username, 40).then(a => {
```
andrew-sunada: mapGetters

src/components/ui/TopBar.vue
```diff
@@ -122,6 +133,9 @@ export default {
     };
   },
   computed: {
+    loggedIn() {
+      return this.$store.state.scope.isLoggedIn;
```
andrew-sunada: We should be using mapGetters here

src/components/ui/TopBar.vue
```diff
@@ -1,15 +1,20 @@
...
     <div class="top-bar__bar">
-      <RouterLink :to="{ name: 'Home' }" class="top-bar__link top-bar__link--logo">
+      <RouterLink
+        :to="{ name: 'Home' }"
+        class="top-bar__link top-bar__link--logo"
+      >
         <img src="@/assets/images/mozilla.svg" :alt="fluent('mozilla')" />
       </RouterLink>
       <SearchForm
+        v-if="loggedIn"
```
andrew-sunada: There is a story for removing the search box if the user is not logged in?

src/components/ui/UserPicture.vue
```diff
@@ -84,15 +84,13 @@ export default {
...
+            this.ownPicture()
           );
         }
       }
     },
     ownPicture() {
-      return (
-        this.avatar.username === this.$store.state.user.primaryUsername.value
-      );
+      return this.avatar.username === this.$store.state.scope.username;
```
andrew-sunada: mapGetters

src/pages/PageHome.vue
```diff
@@ -68,8 +68,11 @@ export default {
     LoadingSpinner,
   },
   computed: {
+    loggedIn() {
+      return this.$store.state.scope.isLoggedIn;
+    },
     username() {
-      return this.$store.state.user.primaryUsername.value;
+      return this.$store.state.scope.username;
```
andrew-sunada: mapGetters

src/pages/PageProfile.vue
```diff
@@ -49,10 +49,7 @@ export default {
   },
   computed: {
     variables() {
-      if (
-        this.$route.params.username ===
-        this.$store.state.user.primaryUsername.value
-      ) {
+      if (this.$route.params.username === this.$store.state.scope.username) {
```
andrew-sunada: mapGetters

fiji-flo:
I see you point. However `scopeV2` is not yet used and does not update after changing data. It also put's a dependency on `scopeV2` to having a public landing page.
90% of these are just renaming. I can refactor the scope store (more likely scope and user) to cater for this if there's a strong preference to do this now.

andrew-sunada:
> @andrew-sunada I see you point. However `scopeV2` is not yet used and does not update after changing data. It also put's a dependency on `scopeV2` to having a public landing page.
> 90% of these are just renaming. I can refactor the scope store (more likely scope and user) to cater for this if there's a strong preference to do this now.

That makes sense. I can approve that for now and create a story to refactor this in the future. There are still 2 comments that have not been addressed not related to mapGetters

src/assets/js/scope.js
fiji-flo: In general I agree. In this case I'd argue that this should be encapsulated in a single place. Making this a constant in this file which in return is only referenced once, again in this file, doesn't feel right.
Anyhow, super weak opinion: if you prefer me moving that out I'm fine. Not sure though what a config file should look like. That would be a simple .js file with constants I guess since all of this is static.

src/components/ui/TopBar.vue
fiji-flo: As per today's call: hiding the search bar for now is part of gradually shipping this.

src/assets/js/scope.js
andrew-sunada: No it's fine. I also prefer a source of truth, but this is just piggy backing off an issue that already existed. I'm fine getting this out so long as we create stories around fixing this.

andrew-sunada: https://jira.mozilla.com/browse/IAM-541

PR #574, opened 2020-03-13
Feature/create

andrew-sunada:
Jira Story: (IAM-458)[https://jira.mozilla.com/browse/IAM-458]

Updating create page to look like mockups


src/components/ui/RadioSelect.vue
```diff
@@ -0,0 +1,113 @@
+<template>
+  <div class="radio-select">
+    <div
+      class="radio-select__option"
+      v-for="(radio, idx) in options"
+      :key="idx"
+    >
+      <input
+        type="checkbox"
```
fiji-flo: I see that it's some work but doing this with an actual radio button would be cool.
For now if fixable I'd suggest moving the `@click` to the div (which enables selecting by clicking anywhere on the div).

More important: it's not working with keyboard navigation. It might be cheaper to use radio buttons and style them (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio) should work nicely with Vue.

src/components/access_group/AccessGroupCreate.vue
```diff
@@ -147,6 +248,7 @@ export default {
         name: this.groupName,
         type: this.groupType,
         description: this.groupDescription,
+        group_expiration: this.selectedExpiration,
```
fiji-flo:
```suggestion
        group_expiration: Number(this.selectedExpiration),
```

andrew-sunada: I don't understand where this suggestion comes from. It already sends the group_expiration as a number.

PR #577, opened 2020-03-16
Fix issue with broken hot reloading

april:


april:
With @fiji-flo's addon, most requests go to localhost except the ones for hot reloading. Those go to IP addresses and they generally don't work. This forces everything to use localhost.

PR #579, opened 2020-03-19
Feature/expiration update

andrew-sunada:
Jira stories: 
 - https://jira.mozilla.com/browse/IAM-521
 - https://jira.mozilla.com/browse/IAM-520

Fixing expiration display

### DIFF
--- a/src/components/access_group/AccessGroupInvitationsEdit.vue
+++ b/src/components/access_group/AccessGroupInvitationsEdit.vue
@@ -69,26 +69,17 @@
               }}
             </div>
           </div>
-          <div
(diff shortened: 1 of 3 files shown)


src/components/access_group/AccessGroupInvitationsEdit.vue
```diff
@@ -362,15 +355,20 @@ export default {
   color: var(--gray-40);
 }
 
-.new-invites-expiration {
+/* .new-invites-expiration {
```
fiji-flo: should be cleaned up in follow up commit.

PR #581, opened 2020-03-23
feat(IAM-430): adding icon to navigate to access group index page

andrew-sunada:
Jira Story: https://jira.mozilla.com/browse/IAM-430

Adding access group icon


src/components/ui/TopBar.vue
```diff
@@ -36,6 +36,15 @@
           <Icon id="search" :width="20" :height="20" />
         </template>
       </ShowMore>
+      <RouterLink
+        v-if="scope.isStaff"
```
fiji-flo: that should be visible to NDA'd.

andrew-sunada: Good call.

PR #583, opened 2020-03-23
Fix/create bug

andrew-sunada:
Jira Story: https://jira.mozilla.com/browse/IAM-459

Fixing create bug
Refactoring Toast
Adding error state to bug

### DIFF
--- a/src/assets/js/fetcher.js
+++ b/src/assets/js/fetcher.js
@@ -52,14 +52,14 @@ class Fetcher {
         },
         body: JSON.stringify(data),
       });
-      return res.status;
-    } catch (e) {
-      console.log('error: ', e);
-      if (this.failOnError && this.isError(e)) {
-        console.log(`got error → failing over`);
-        // reload();
+      const responseObject = await res.json();
+      if (res.status >= 200 && res.status < 300) {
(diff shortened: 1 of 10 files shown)


src/components/access_group/AccessGroupCreate.vue
```diff
@@ -268,7 +273,7 @@ export default {
 };
 </script>
 
-<style>
+<style scoped>
```
fiji-flo: We should be consistent there in the future. We start to have a mix. A follow up thing.

src/assets/js/fetcher.js
```diff
@@ -52,14 +52,14 @@ class Fetcher {
         },
         body: JSON.stringify(data),
       });
-      return res.status;
-    } catch (e) {
-      console.log('error: ', e);
-      if (this.failOnError && this.isError(e)) {
-        console.log(`got error → failing over`);
-        // reload();
+      const responseObject = await res.json();
```
fiji-flo: This will display generic network errors. This easily happens when deleting the pmo-access cookie while on the /a/create page and then trying to submit.

I filed a ticket and we'll solve it later.

PR #585, opened 2020-03-27
Fix/wrap up

andrew-sunada:
Wrapping up all primary issues for access group.

- Fixing "edit members expiration" area styles
- Fixing members table with "no expiration" to be a dash
- Fixing "load more" button styles
- Linking "load more" action to api "next" field


andrew-sunada:
The pagination on the members table is currently broken for lack of understanding of how pagination works @fiji-flo . Please advise and we shall fix this ;)

src/components/_mixins/MembersListMixin.vue
```diff
@@ -0,0 +1,100 @@
...
+      this.getMembersWithOptions({
+        groupName: this.groupName,
+        options: this.membersListOptions,
+      });
+    },
+    resetOptions() {
+      this.membersListOptions = defaultOptions;
+    },
+  },
+  data() {
```
fiji-flo: That `data` is only instantiated once. The Mixin is shared across all usages. So either we reset it from within the component where it's used, or we store the state somewhere else (Vuex?).

src/components/_mixins/MembersListMixin.vue
```diff
@@ -27,6 +27,7 @@ export default {
   },
   mounted() {
     this.fetchMembers(this.groupName);
+    this.membersListOptions = defaultOptions;
```
fiji-flo, line 31: This didn't fix it. I sent you a video on slack. Digging deeper, I think the issue is that `defaultOptions` actually gets changed. To observe just add a `console.log(defaultOptions);` here.

src/components/access_group/AccessGroupMembersEdit.vue
```diff
@@ -200,8 +207,7 @@ export default {
     RadioSelect,
     Select,
   },
-  props: [],
-  mounted() {},
+  mixins: [MembersListMixin],
```
fiji-flo, line 210: I just saw this file exports the wrong name `AccessGroupInformationEdit` instead of `AccessGroupMembersEdit`

src/components/_mixins/MembersListMixin.vue
andrew-sunada: Good call. fixed that.

src/components/access_group/AccessGroupMembersEdit.vue
andrew-sunada: 👀 ...nothing to see here folks...

src/components/_mixins/MembersListMixin.vue
```diff
@@ -28,7 +29,7 @@ export default {
   },
   mounted() {
     this.fetchMembers(this.groupName);
-    this.membersListOptions = defaultOptions;
+    this.membersListOptions = getDefaultOptions();
```
fiji-flo, line 32: That's not needed anymore if I see this right sorry for the fake news earlier.

PR #586, opened 2020-03-30
Public page

fiji-flo:


src/assets/images/skewed-triangle.svg
```diff
@@ -0,0 +1,13 @@
...
+   xmlns="http://www.w3.org/2000/svg"
+   height="101.17517"
+   width="92.936852"
+   version="1.1"
+   viewBox="0 0 92.936852 101.17517">
+  <polygon
+     transform="rotate(-156,39.752044,60.969868)"
+     points="100,100 0,100 50,13.397 "
+     id="tri"
+     style="fill:none;stroke:#0060DF;stroke-opacity:1;stroke-width:0.5" />
```
andrew-sunada: might wanna set the stroke to "currentColor"?

src/components/ui/Footer.vue
```diff
@@ -1,43 +1,67 @@
...
-    <select v-if="languages.length > 1" v-on:change="updateLocale">
-      <option
-        v-for="[locale, language] in languages"
-        :key="locale"
-        :value="locale"
-        :selected="locale === currentLocale ? true : false"
-        >{{ language }}</option
+    <section class="footer__contribute" v-if="showContribute">
+      <h1>{{ fluent('contribute_header') }}</h1>
+      <div class="footer__contribute-img">
```
andrew-sunada: I'm concerned with this class name as it doesn't seem follow BEM standards. Not saying that we HAVE to do that, but I'd like to work with that. If you think otherwise, that's fine, but I think it's worth discussion.

src/components/ui/Footer.vue
```diff
@@ -1,43 +1,67 @@
...
+        <img
+          src="@/assets/images/crowd.png"
+          srcset="@/assets/images/crowd@2x.png 2x"
+        />
+      </div>
+      <p>{{ fluent('contribute_text') }}</p>
+      <ExternalButtonLink
+        class="footer__contribute__link button--invert"
```
andrew-sunada: This class is confusing to me too regarding BEM standards.

src/components/ui/Footer.vue
```diff
@@ -1,43 +1,67 @@
...
 export default {
   name: 'Footer',
   mixins: [LinksMixin],
   components: {
     Icon,
+    ExternalButtonLink,
   },
   computed: {
+    showContribute() {
+      return !this.$store.state.scope.isLoggedIn && this.$route.name === 'Home';
```
andrew-sunada: Should we put this "Home" string in a static variable?

src/components/ui/Footer.vue
```diff
@@ -89,10 +113,58 @@ export default {
...
+.footer__contribute {
+  padding: 2em 2em 0em;
+  background-color: var(--white);
+  display: grid;
+  grid-template-columns: 1fr;
+}
+.footer__contribute > h1 {
+  font-size: 2.5em;
+}
+.footer__contribute__link {
```
andrew-sunada: Thoughts on class name vs BEM?

src/pages/PageHome.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <main class="home">
-    <div class="home__intro">
+    <div v-if="loggedIn" class="home__intro home__intro_23">
```
andrew-sunada: I'd also love to discuss this class name "home__intro_23".

src/assets/images/skewed-triangle.svg
fiji-flo: `currentColor` does not work in an non-embedded svg (see the other svgs in that folder).

src/components/ui/Footer.vue
fiji-flo: Trying to fix all of those. Thanks. Still not doing this intuitively.

src/components/ui/Footer.vue
fiji-flo: We have that in a couple of places. I'd do that in a follow up as it's a general refactoring and then we should do that with all the old routes.

src/components/ui/ExternalButtonLink.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <a class="button button_link" :href="href">
```
andrew-sunada, line 2: I don't know if 'button_link' was copy pasted or something, but that schema doesn't fit in BEM. Revisit?

src/components/ui/ExternalButtonLink.vue
```diff
@@ -0,0 +1,39 @@
...
+    href: String,
+    text: String,
+    iconLeft: String,
+    iconRight: String,
+  },
+};
+</script>
+
+<style>
+.button--link {
```
andrew-sunada, line 25: maybe consider putting a comment about link here?

PR #590, opened 2020-04-06
Add volunteer slack-access classification

### LINKED ISSUE #530: Add slack-access data classification
**Is your feature request related to a problem? Please describe.**
Checking whether a contributor has access to the Slack instance is always complicated, as you need to scroll to the very bottom and then check whether the user has the `slack-access` access group.

**Describe the solution you'd like**
Like with the current markers, I'd imagine something next to it that appears as soon as this user has access to Slack.

See this:
image showing a Slack badge next to the `Mozilla-Confidential` badge

**Describe alternatives you've considered**
Maybe this could be expanded to a popover of the user in other contexts as well.

ref #508 

---

If wanted, this seems like a thing I could contribute myself. :)

jens1o:
It is often handy to see at glance whether a contributor has Slack access
for further communication and information sharing.

image

Fixes #530


src/components/profile/DataClassification.vue
```diff
@@ -69,12 +91,26 @@ export default {
...
       }
     },
+    isVolunteerWithSlackAccess() {
+      return (
+        !this.staffInformation.staff.value &&
+        'slack-access' in (this.accessInformation.mozilliansorg.values || {})
+      );
+    },
+    slackLabel() {
+      return 'Slack';
```
jens1o, line 101: I seemed urge to add a function for that because the `pre-commit` hook `husky` was denying a hardcoded span label at all cost... And I didn't think this is something which can be localized.

src/components/profile/DataClassification.vue
```diff
@@ -1,24 +1,46 @@
...
-        :tags="{
-          wiki: {
-            tag: 'a',
-            target: '_blank',
-            href: 'https://wiki.mozilla.org/Security/Data_Classification',
-          },
-        }"
-      />
-    </Tooltip>
+  <div class="data-classification-list">
```
fiji-flo, line 2: I propose we refactor this into an `AccessLabel` component. Which can be configured to display data access or slack access for now. This component would then be added twice into `ViewPersonalInfo`.

src/components/profile/DataClassification.vue
fiji-flo: I see. In that case it should live in there as data (move to it the `data()` method / create one here).

src/components/profile/DataClassification.vue
```diff
@@ -95,7 +131,13 @@ export default {
...
 }
 
-.data-classification__label {
+.data-slack__access {
+  background-color: #611f69;
+  color: #fff;
+}
+
+.data-classification__label,
+.data-slack__access__label {
```
fiji-flo, line 140: In terms of CSS naming we try to follow http://getbem.com/naming/ but these should change if we refactor that  component anyhow.

jens1o:
I'm not 100% sure whether I understood the BEM guide right, hence I'm looking forward to see feedback on that!

src/components/profile/AccessLabel.vue
```diff
@@ -69,12 +101,23 @@ export default {
...
 </script>
 
 <style>
-.data-classification {
+.data-access-label:not(:last-child) {
+  margin-right: 0.3em;
+}
+
+.data-classification,
+.data-slack-access {
```
fiji-flo, line 120: I'd rename this to something like `access-label` (which make the above rule simpler, too / you don't need an extra class anymore).
And then have the modifiers `access-label--data-staff`, `acces-label--data-confidential`, `access-label--slack`.

jens1o: Just asking: With or without the `data-` class-name prefix? I'm unsure about this.

fiji-flo: this specific class I would replace with just on class called `access-label` without `data-`

PR #597, opened 2020-04-15
Public fixes

fiji-flo:


src/components/access_group/AccessGroupView.vue
```diff
@@ -124,7 +124,7 @@ export default {
   margin-bottom: 0;
 }
 
-@media (min-width: 50em) {
+@media (min-width: 57.5em) {
```
jens1o, line 127: Could you describe how you found out these values (~800px -> ~920px when 1em = 16px)?

fiji-flo: In this case I just wanted to consolidate to one break point. We had 50em and 57.5em mixed.

fiji-flo: Backed it out again. I wanted to see how this behaves. But this would make the tablet experience worse. I'll try to improve this in another way.

fiji-flo:
Merging to get matt's input. Happy to reopen this. This are only deletions and CSS.

PR #608, opened 2020-05-06
Onboarding

fiji-flo:


src/components/guide/OnboardingModal.vue
```diff
@@ -0,0 +1,147 @@
...
+        this.step_data = this.steps[this.step - 1];
+      }
+    },
+    done() {
+      this.$store.state.onboarding.modal = false;
+      this.$emit('close');
+    },
+  },
+  mounted() {
+    this.swipe = new Swipe(this.$el, this.next, this.prev);
```
jens1o, line 79: Bringing this up out of curiosity: In Right-to-left regions, do they also swipe to the right to go to the next tile normally? I just wonder whether this is a native experience for them as well. :)

fiji-flo: Interesting question. That would affect the forward / backward button in FF, too?

PR #614, opened 2020-05-18
Feat/dropdown

andrew-sunada:
Fixing auto complete in tag selector; both styles and accessibilty: IAM-663

### DIFF
--- a/src/assets/js/access-groups-api.config.js
+++ b/src/assets/js/access-groups-api.config.js
@@ -141,7 +141,13 @@ export default {
   },
   users: {
     endpoint: `${API_PREFIX}users`,
-    get: (endpoint, q, groupName, includeCurators = false) =>
-      `${endpoint}?q=${q}&g=${groupName}&c=${includeCurators}`,
+    get: (
+      endpoint,
+      q,
+      groupName,
+      includeCurators = false,
+      showExisting = false,
+    ) =>
+      `${endpoint}?q=${q}&g=${groupName}&c=${includeCurators}&a=${showExisting}`,
   },
 };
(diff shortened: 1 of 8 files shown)


andrew-sunada:
> Hey, looks nice. The same should be used in the curators list.
> 
> In terms of accessibility there's much more to it than the up/down arrow keys: https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
> I'd still recommend to drive the Combobox component for this. **But please at least support selecting via and closing via . Currently will submit the form (which it should if the input is being in focus).**

I'm confused by the last part of the comment ^

andrew-sunada:
> > Hey, looks nice. The same should be used in the curators list.
> > In terms of accessibility there's much more to it than the up/down arrow keys: https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html
> > I'd still recommend to drive the Combobox component for this. **But please at least support selecting via and closing via . Currently will submit the form (which it should if the input is being in focus).**
> 
> I'm confused by the last part of the comment ^

I went down this road trying to incorporate the combobox but there's so many differences and the code is so unreadable that I spent 3 hours of work on it unsuccessfully. The main difference being injecting a vue component into the preact system. The further I went into the more hacky it felt. 

I'm willing to take a break and hit it again, but at the rate that we're trying to solve this before you take off, I think i'd rather stick with the solution I have. 

You said "But please at least support selecting via and closing via" What does that mean? I can probably work on getting that supported by getting this worked through super early i the morning so you can confirm it.

fiji-flo:
esc and enter. Sorry markdown swallowed these (don't ask why I thought putting them into \<\> is helpful).

andrew-sunada:
> esc and enter. Sorry markdown swallowed these (don't ask why I thought putting them into <> is helpful).

Cool. I've updated the escape and enter fields

PR #619, opened 2020-08-11
Introduce usage of srcset for avatars

jens1o:
supersedes #451


src/components/ui/UserMenu.vue
```diff
@@ -11,12 +11,14 @@
           }"
           :size="40"
           :pictureSize="100"
```
fiji-flo, line 13: We can finally remove `pictureSize` with this

PR #620, opened 2020-08-11
Introduce skeleton preview of profile page

jens1o:
supersedes (and heavily based on) #545


jens1o:
If there is any logic error or thing that I missed, I take that. Was a not so pleasant time working on this…

PR #623, opened 2020-08-12
Make adding tags more user-friendly

jens1o:
fixes https://github.com/mozilla-iam/dino-park-issues/issues/177
fixes https://github.com/mozilla-iam/dino-park-issues/issues/179

addresses https://github.com/mozilla-iam/dino-park-issues/issues/134#issuecomment-548298112

### DIFF
--- a/src/components/profile/edit/EditTags.vue
+++ b/src/components/profile/edit/EditTags.vue
@@ -121,14 +121,14 @@ export default {
     },
     handlePaste() {
       this.$nextTick(() => {
-        let text = this.$refs.inputTag.value.trim();
+        let text = this.$refs.inputTag.value;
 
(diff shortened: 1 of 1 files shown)


src/components/profile/edit/EditTags.vue
```diff
@@ -117,6 +119,19 @@ export default {
         this.tags.values.splice(index, 1);
       }
     },
+    handlePaste() {
+      this.$nextTick(() => {
+        let text = this.$refs.inputTag.value.trim();
```
fiji-flo, line 124: Maybe we should support just `,` as separator and add do something like
```js
for (const tag of text.split(',')) {
  this.addTag(tag.trim());
}
```

PR #624, opened 2020-08-13
Simplify access group list, decouple it from the vuex store

jens1o:


src/components/access_group/AccessGroupList.vue
```diff
@@ -45,13 +45,17 @@ import SearchForm from '@/components/ui/SearchForm.vue';
...
 import Icon from '@/components/ui/Icon.vue';
+import { Api } from '@/assets/js/access-groups-api.js';
+import { AbbGroupViewModel } from '@/view_models/AccessGroupViewModel';
 
 const resultsStep = 20;
 const defaultListOptions = {
   search: '',
   sort: 'member-count-desc',
   numResults: resultsStep,
+  next: void 0,
```
fiji-flo, line 56: I think `null` is fine here. Otherwise please use `undefined`. AFAIK `void 0` comes from the dark age of JavaScript.

jens1o: No, `void 0` is really modern, as you can theoretically overwrite `undefined` and use it as a variable name: https://stackoverflow.com/a/5717084/8496913

Anyway, I'll test `null` and use it here instead.

jens1o: Ah, I remember why I did this. When `defaultListOptions.ǹext` is null, `listOptions` don't get overwritten to a falsey value (just noticed that during final testing)... I'll revert the change back to be `void 0`.

fiji-flo (approved):
:+1:

PR #625, opened 2020-08-14
Simplify member list of access groups

jens1o:

PR #626, opened 2020-08-26
AccessGroup: Get rid of vuex store mechanism

jens1o:
replaces #625


fiji-flo:
There are more events that modify Group date (basically every POST/PUT/DELETE). Make sure they emit the event.

fiji-flo (approved):
Looks great. Let's move this to testing!

PR #659, opened 2021-06-02
Adding docker env variables to codebuild

duallain:
rtucker wants to login to dockerhub to prevent api rate limit errors.
This change makes it so an ssm parameter store secret is available for the pipelines.


rtucker-mozilla (approved):
Looks great

PR #662, opened 2021-09-09
Adding ghe_group_curators to list of NDA'd groups

rtucker-mozilla:
Adding ghe_group_curators to list of NDA'd groups

### DIFF
--- a/src/assets/js/scope.js
+++ b/src/assets/js/scope.js
@@ -28,6 +28,7 @@ class Scope {
     this.isStaff = isStaff;
     this.isNdaed =
       'nda' in (mozilliansorgGroups || {}) ||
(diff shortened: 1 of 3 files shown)


src/assets/js/scope.js
```diff
@@ -28,6 +28,7 @@ class Scope {
     this.isStaff = isStaff;
     this.isNdaed =
       'nda' in (mozilliansorgGroups || {}) ||
```
floatingatoll, line 30: Could you please add a code comment pointing to the _other_ locations here, too, since we seem to have hard-coded this in so many places?

I will not be sad if you choose not to, and instead have each of the pull requests reference the others. That would be fine too instead. Since this will always be spelunking.

floatingatoll (approved):
r+ with comments

PR #669, opened 2022-03-30
remove dn from profile views

floatingatoll:
Per discussion (see email). I'd like to have this deployed to whatever our "non prod" is so that I can get before/after screenshots for internal comms, and confirm that this doesn't break anything. Please hold prod until coordinated.

OK to merge as soon as you think it's a good time for that; I'm not up-to-date on the raw semantics of this particular IAM repository, so whatever's best is fine with me.


bkochendorfer (approved):

floatingatoll:
Resigned and merged.

PR #673, opened 2022-05-09
feature(onboarding): Add error messaging 

bkochendorfer:
Currently if a user has a profile in the CIS database but is not set to active we will continuously query CIS, eventually giving up but not showing any error to the user. This updates the frontend to show an error to the user and where they can go for help.

Screenshot 2022-05-09 at 14-02-11 Mozilla People Directory

### DIFF
--- a/src/locales/en-US/strings.ftl
+++ b/src/locales/en-US/strings.ftl
@@ -109,7 +109,7 @@ onboarding_modal_username = Getting started
   .username_chars = Username must only contain latin characters a-z or digits 0-9
 
 onboarding_modal_profile_error =
-  .profile_error = Error in creating profile please reach out in <a data-l10n-name="iam_slack">#iam</a> Slack or at <a data-l10n-name="support_email">people-mozilla-org-support@mozilla.com</a>
+  .profile_error = Error in creating profile. Please reach out in <a data-l10n-name="iam_slack">#iam</a> Slack or at <a data-l10n-name="support_email">people-mozilla-org-support@mozilla.com</a>
 
 tooltip_tour = Welcome
   .skip = Skip tour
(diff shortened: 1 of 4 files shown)


src/store/index.js
```diff
@@ -45,6 +47,8 @@ async function fetchUser(commit) {
 async function retryFetchUser(commit, retries = 120) {
   if (!(await fetchUser(commit)) && retries > 0) {
     setTimeout(() => retryFetchUser(commit, retries - 1), 1000);
+  } else if (!userFound) {
```
bkochendorfer, line 50: Not sure if this is a better approach here, I am not used to front end or Vue! Let me know

src/locales/en-US/strings.ftl
```diff
@@ -108,6 +108,9 @@ onboarding_modal_username = Getting started
   .username_space = Username must not contain spaces
   .username_chars = Username must only contain latin characters a-z or digits 0-9
 
+onboarding_modal_profile_error =
+  .profile_error = Error in creating profile please reach out in <a data-l10n-name="iam_slack">#iam</a> Slack or at <a data-l10n-name="support_email">people-mozilla-org-support@mozilla.com</a>
```
smarnach, line 112: I feel these are two sentences, so their should be some interpunction before "please" – probably a peiod.

smarnach (approved):

src/locales/en-US/strings.ftl
bkochendorfer: :+1: Good idea!

floatingatoll:
(It's too late for me to register a change request, so I'll leave a comment.)

Please update the error message to provide instructions on what to say _in_ chat/email; for example, "and say inactive profile error in your message" or whatever seems best.

bkochendorfer:
@floatingatoll oops, i'm still testing this in the test environment. If you want to put in a PR to the message I'll make sure to get it out

floatingatoll:
No it's okay! I'm so happy to see development and repair occurring! Don't hold the release on me, this is going to be a rare case, I'll provide a PR when able.

PR #696, opened 2023-02-02
apply the view costCenter filter to edit

floatingatoll:
Something in the CIS pipeline is appending ".0" to cost centers, which is then displayed to user _only_ when editing their profiles, not when viewing those of others.

This patch applies the logic from the view page to the (readonly) field on the edit page as well, so that we're internally consistent within Dinopark and also consistent with the current 2023-era cost centers as shown in Finance documents.

We still need to find what's injecting .0 and fix that, but this removes the user-visible impact for Dinopark until we do so, and will noop once that's removed upstream.

### DIFF
--- a/src/components/profile/edit/EditPersonalInfo.vue
+++ b/src/components/profile/edit/EditPersonalInfo.vue
@@ -354,7 +354,10 @@
           type="text"
           id="field-cost-center"
           disabled
-          :value="staffInformation.costCenter.value"
+          :value="
+            staffInformation.costCenter.value &&
+            staffInformation.costCenter.value.replace(/\.0$/, '')
+          "
         />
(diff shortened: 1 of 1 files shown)


gcoxmoz (approved):
Hrm.
I do see the described issue when I edit my profile.  I do see this as scoped to the right area of the codebase.
I don't know the language.  I don't know the strong-type-iness in the language.  I don't know that it's a string, like, maybe it's presenting as a float and maybe this is going to blow up.

This is a pro-forma approval ala "I have no better offer and it looks like it makes sense, let's find out"  If it blows up, it wouldn't surprise me.  But, it's low-stakes and no big deal on a revert, so let's go.

PR #706, opened 2024-02-12
IAM-1294 Remove notifications link in people.m.o dropdown menu

frosemond:
- removed user notifications menu option and associated strings 
- removed manage auto login menu option and associated strings


dividehex (approved):
👍

